Issue #20, opened 2026-05-17
Define machine-readable diagnostics output contract


ss-o:
## Summary

Define the first stable machine-readable output format for diagnostics and parser failures.

## Why this matters

CI and editor integrations need a predictable transport format, but the project should define it once on top of the shared diagnostics model rather than letting each future consumer invent its own shape.

## Suggested project triage

- Type: Feature
- Priority: High
- Effort: Medium
- Status: Ready

## Acceptance criteria

- A versioned machine-readable schema is documented.
- Diagnostics include rule ID, severity, message, and source range.
- Parser failures and analyzer diagnostics have an explicit representation strategy.
- The contract is covered by tests once implemented.

ss-o:
## Dependency note (2026-06-01)

The shared prerequisite is now on `main`: `internal/diag` provides renderer-agnostic diagnostics with stable rule IDs, severity, messages, source ranges, deterministic sorting, and tests.

This issue remains open for the transport contract itself. Schedule it after corpus workflow `#6/#9/#10`, first-wave rule policy `#7`, and suppression contract `#19`, so JSON or other machine-readable output is defined on top of stable analyzer semantics rather than in parallel with them.

ss-o:
## Agent handoff (2026-06-12)

PR #57 implements this issue now that the prerequisites are merged (#54 corpus, #55 rule policy, #56 suppression contract):

- `docs/project/output-contract.md` — versioned envelope contract (v1) with versioning policy.
- `internal/diag.WriteJSON` + tests — rule ID, severity, message, half-open source range; deterministic ordering; `diagnostics` never null.
- `zsh-lint --format=json` — opt-in; human output and exit codes unchanged; parser failures represented as reserved `parse/error` diagnostics with zero-width ranges.
- Suppression interaction defined per `docs/project/suppression.md`: suppressed findings omitted, `meta/*` always included (no version bump needed when the analyzer starts emitting them).

ss-o:
Implemented by merged PR #57: `docs/project/output-contract.md` (versioned v1 envelope), `internal/diag.WriteJSON` with deterministic ordering, and `zsh-lint --format=json` with parse failures as reserved `parse/error` diagnostics. Acceptance criteria met; closing.

Issue #19, opened 2026-05-17
Define inline suppression model and syntax


ss-o:
## Summary

Define how users suppress intentional findings without weakening unrelated diagnostics.

## Why this matters

The first-wave rule policy requires every shippable rule to document suppression behavior. The analyzer needs one shared suppression contract before rule work begins so later diagnostics do not invent incompatible escape hatches.

## Suggested project triage

- Type: Architecture
- Priority: High
- Effort: Medium
- Status: Ready

## Acceptance criteria

- Inline suppression syntax is documented.
- Suppressions can target a specific rule ID.
- Scope semantics are explicit.
- Malformed or stale suppressions have a defined handling strategy.

ss-o:
## Agent handoff (2026-06-12)

PR #56 defines the suppression contract in `docs/project/suppression.md`:

- Syntax: `# zsh-lint disable=<rule-id>[,...] [-- reason]` — rule IDs required, no blanket disable.
- Scopes: trailing (same line) and preceding (next source line) only; file/block scope deliberately out of the first wave.
- Malformed directives → `meta/malformed-suppression` (warning), suppress nothing.
- Stale/unknown-ID suppressions → `meta/unused-suppression` (info).
- The #20 machine-readable contract must represent suppression and `meta/*` diagnostics explicitly (requirements listed in the doc).

Implementation (directive parsing in the analyzer + the two `meta/*` rules) is a follow-up once this contract is reviewed and merged. Feasibility verified: the front end keeps comments (`syntax.KeepComments(true)`).

ss-o:
Contract defined in #56 (merged to `next` at 95b7186): `docs/project/suppression.md`. Acceptance criteria met — syntax ✔, rule-ID targeting ✔, scope semantics ✔, malformed/stale handling ✔. Analyzer implementation (directive parsing + `meta/malformed-suppression` and `meta/unused-suppression`) is follow-up work and must cite this contract.

Issue #17, opened 2026-05-17
Compare mvdan/sh against active tree-sitter-zsh frontend on the survey corpus


ss-o:
## Summary

Run the same documented Z-Shell survey corpus through the active `georgeharker/tree-sitter-zsh` grammar and compare the result with the current `mvdan/sh` frontend before making any long-term frontend commitment beyond the reboot phase.

## Why this matters

`mvdan/sh` already gives the project a maintained Go parser and useful AST foundation, but the first real-world survey exposed meaningful Zsh gaps. An active `tree-sitter-zsh` grammar now exists and should be measured rather than ignored, even if it is not yet the recommended core frontend.

## Suggested project triage

- Type: Research
- Priority: High
- Effort: Medium
- Status: Ready

## Acceptance criteria

- The same 16-file survey corpus is evaluated with the active tree-sitter grammar.
- Results are compared fixture-by-fixture against the current frontend.
- The tradeoff between AST usefulness, parser fidelity, and implementation cost is documented.
- The comparison informs a future ADR rather than relying on assumptions.

ss-o:
First comparison pass completed against the same 16-file survey corpus.

Result:

| Frontend | Parsed | Failed |
| --- | ---: | ---: |
| `mvdan/sh` v3.13.1 | 7 | 9 |
| `georgeharker/tree-sitter-zsh` @ `86b37f8` | 6 | 10 |

Notable difference:
- tree-sitter parses `src/public/zsh/init.zsh` with its documented `always` block
- but it fails on `zd/docker/utils.zsh` and `zunit/build.zsh`, which `mvdan/sh` parses successfully

Interpretation: tree-sitter is worth tracking and may be useful later, but the first survey does **not** justify an immediate frontend switch. The detailed comparison is now recorded in `docs/project/2026-05-17-frontend-comparison.md`.

ss-o:
Evidence update (2026-06-12, PR #58): on the documented 19-file corpus, mvdan/sh LangZsh parses 11 (13 on master preview) vs tree-sitter-zsh's 6/16 from the earlier run. The dedicated Zsh dialect upstream removes the main motivation for a frontend swap; suggest keeping tree-sitter as tracking-only and noting the LangZsh numbers when the comparison doc (still uncommitted) is written.

ss-o:
[...]

Issue #8, opened 2026-05-17
Evaluate parser coverage against real Z-Shell code and derive first rule candidates


ss-o:
## Summary

Run the rebooted parser frontend against representative real-world Z-Shell repositories, record parser gaps, and use the resulting evidence to shape the first trustworthy semantic-rule candidates.

## Why this matters

The project should choose its first diagnostics from real code and parser behavior, not from intuition. This phase turns the reboot foundation into an evidence base for the first analyzer milestone.

## Suggested project triage

- Type: Research
- Priority: High
- Effort: Large
- Status: Ready

## Scope

- select the initial Z-Shell corpus repositories / files
- run `mvdan/sh` Zsh parsing across representative samples
- minimize and record parser gaps as they appear
- identify recurring semantic-risk patterns worth considering for first-wave rules
- produce a written recommendation for the initial 15-25 rule candidates

## Acceptance criteria

- Initial corpus selection is documented.
- Parser successes and failures are measured against representative real code.
- Any parser gaps are tracked as dedicated issues or linked findings.
- Candidate first-wave rules are ranked by confidence and value.
- The findings are captured in project docs so future work can resume without repeating the survey.

ss-o:
## Roadmap reconciliation (2026-06-01)

This issue is partially complete on current `main`.

### Landed foundation
- `cmd/zsh-lint-survey` is a dedicated parser-gap survey command.
- `internal/survey` emits greppable parser-failure diagnostics and a summary.
- The initial real-code findings were split into dedicated parser-gap follow-ups `#11`-`#17`.
- `internal/diag` is the shared renderer-agnostic diagnostics model.
- Five initial semantic rules now exist under `internal/rules`.

### Remaining work
1. Complete the reusable corpus and parser-gap contribution workflow in `#6`, `#9`, and `#10`.
2. Use that documented corpus workflow to rerun the survey and refresh the candidate ranking required by this issue.
3. Complete first-wave rule policy in `#7` before expanding the rule set.
4. Compare the active tree-sitter frontend in `#17` before making a longer-term parser commitment.

Keeping this issue open as the research umbrella; implementation slices remain in the linked issues.

ss-o:
## Agent handoff — survey refresh (2026-06-12)

PR #54 reruns the survey over the now-documented corpus (`docs/project/corpus.md`, 19 files):

| Front end | Parsed | Failed |
| --- | ---: | ---: |
[...]

PR #40, opened 2026-05-29
feat(diag): diagnostics model (internal/diag)

ss-o:
Implements the diagnostics model from the design spec (zsh-lint#18, Linear ZSH-3).

Adds a parser-neutral, renderer-agnostic `internal/diag` package:
- `Position`/`Range` — own value types, **no** `mvdan/sh` dependency (preserves front-end swappability, #17)
- `Severity` enum (Error/Warning/Info/Hint, LSP order) with tested `String`/`ParseSeverity`
- opaque `RuleID` (slug convention) + pure-data `Diagnostic`
- `Diagnostics` with one fully input-independent `Sort` (File → Range.Start → RuleID → Severity → Message)

No rules (#8), suppressions (#19), or output formatting (#20) — only the shared model those build on.

Built via brainstorm → spec → plan → subagent-driven TDD, with spec-compliance and code-quality review gates.

## Test plan
- [x] `go build/vet/test ./...` green
- [x] `internal/diag` unit tests: severity round-trip + unknown-input error, position/range validity, sort determinism incl. unpositioned-first and full tiebreak chain
- [x] gofmt clean


ss-o:
Addressed all review feedback in a0eaca7:
- `Sort` now also tiebreaks on `Range.End` (Line, Column, Offset), so order is input-independent for distinct diagnostics; added `TestDiagnosticsSortBreaksTiesOnRangeEnd`.
- Corrected `Position.IsValid`/`Range.IsValid` doc comments to state exactly what they check (Line set / Start set; End and Start<End are not validated).
- Reworded the severity tiebreak test to clarify Error is most-severe / numeric 0 under ascending sort.

Issue #63, opened 2026-06-12
Rule proposal: plugin/function-scoped-options


ss-o:
## Rule ID

`plugin/function-scoped-options`

## Rule name

Function files should scope shell options

## Summary

Reports autoloadable function files (files under a `functions/` directory) that neither run `builtin emulate -L zsh` nor `setopt local_options` before executing option-sensitive statements, so their behavior silently depends on the caller's option state.

## Why this matters

Zsh functions inherit the calling shell's options. An autoloaded widget or handler executed under `sh_word_split`, `ksh_arrays`, or `no_extended_glob` parses and behaves differently than the author tested. The Z-Shell Plugin Standard's recommended-options guidance and the org convention for annex handlers (`builtin emulate -L zsh ${=${options[xtrace]:#off}:+-o xtrace}`) exist precisely for this. Manual grounding: `zshoptions` (LOCAL_OPTIONS), `zshmisc` §Functions, Plugin Standard §Standard Recommended Options.

## Corpus evidence (2026-06-12, LangZsh-parseable files)

Violations (no scoping at all):

- `z-a-meta-plugins/functions/.za-meta-plugins-meta-cmd-help-handler`
- `zsh-fancy-completions/functions/.completion-prediction`
- `zsh-fancy-completions/functions/.force_rehash`

Compliant alternatives (must stay silent):

- `zsh-fancy-completions/functions/.expand-or-complete-with-dots:4` — `setopt typeset_silent local_options`
- `zsh-fancy-completions/functions/.complete_menu:4` — `setopt local_options always_last_prompt`
- `z-a-meta-plugins/functions/.za-meta-plugins-meta-cmd` — strict `builtin emulate -L zsh`

## Bad / Good examples

Bad (`functions/.my-widget`):

```zsh
local -a matches
matches=( $~pattern )   # behavior depends on caller's GLOB/SH_WORD_SPLIT
```

Good:

```zsh
builtin emulate -L zsh
local -a matches
matches=( $~pattern )
```

Also good: `setopt local_options extended_glob` as the first option-touching statement.

## Proposed severity

`hint`

## Severity justification

Per the severity mapping, style/idiom preferences ship at `hint`; the pattern is a latent-bug enabler rather than broken code, and intentional caller-state-mutating functions exist.

## False-positive analysis

1. **Functions that intentionally mutate caller state** (e.g. option-toggling helpers): rule stays at `hint`; intentional cases use suppression with a reason.
[...]

PR #71, opened 2026-06-23
feat(rules): add plugin/function-scoped-options rule

ss-o:
## Summary

Implements the approved rule proposal **`plugin/function-scoped-options`** (#63):
reports autoloadable function files (files under a `functions/` path segment)
that execute logic before localizing inherited shell options.

A file is **compliant** when, before its first non-guard executable top-level
statement, it scopes options via either:

- `emulate -L zsh` (optionally `builtin`-prefixed; grouped switches like `-LR`
  accepted; shell name must be the literal `zsh`), or
- `setopt local_options` (optionally `builtin`-prefixed; Zsh option-name
  normalization — case-insensitive, underscores ignored).

A contiguous prefix of narrow early-exit guards is allowed before scoping:
`condition || return` and single-return `if` (no `elif`/`else`).

Severity: **hint**. One diagnostic per file, at the first offending statement.
Inline suppression via the existing analyzer pass
(`# zsh-lint disable=plugin/function-scoped-options -- <reason>`).

## Notes for review

- The implementation is **broader than the original plan's code sketch** but
  stays within the design doc's stated intent: it also rejects
  `emulate -L zsh -c …` (command mode does not localize the file's options),
  handles `setopt` inverse forms (`+o local_options` / `no_local_options`,
  last-wins), and rejects backgrounded/redirected/negated guard shapes.
- **Corpus validation (issue #63 evidence):** the analyzer flags
  `zsh-fancy-completions/functions/.completion-prediction` and
  `…/.force_rehash`. The originally cited
  `z-a-meta-plugins/functions/.za-meta-plugins-meta-cmd-help-handler` is now a
  fully commented-out stub and is correctly silent; compliant alternatives
  stay silent.

## Verification

`gofmt` clean · `go test ./...` · `go vet ./...` · `go build ./...` ·
gomarkdoc generation includes the new rule.

## Refs

- Implements #63
- Tracker: ZSH-17 (zsh-lint corpus workflow & analyzer contract roadmap)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/rules/function_scoped_options.go
+++ b/internal/rules/function_scoped_options.go
@@ -81,9 +81,7 @@ func isOptionScopingStatement(stmt *syntax.Stmt) bool {
 func hasUnsafeStatementEffect(stmt *syntax.Stmt) bool {
 	return stmt.Background ||
 		stmt.Coprocess ||
-		stmt.Disown ||
-		stmt.Negated ||
-		len(stmt.Redirs) != 0
+		stmt.Disown
 }
 
 func literalCommand(call *syntax.CallExpr) (int, string, bool) {
--- a/internal/rules/function_scoped_options.go
+++ b/internal/rules/function_scoped_options.go
@@ -55,7 +55,7 @@ func hasFunctionsPathSegment(path string) bool {
 }
 
 func isOptionScopingStatement(stmt *syntax.Stmt) bool {
-	if stmt == nil {
+	if stmt == nil || hasUnsafeStatementEffect(stmt) {
 		return false
 	}
 	call, ok := stmt.Cmd.(*syntax.CallExpr)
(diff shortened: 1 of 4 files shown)


ss-o:
## Final review follow-up (2026-07-17)

A fresh pre-merge review found that `return 1 ignored` was incorrectly accepted as a leading early-return guard. In Zsh, that command reports `return: too many arguments` and continues executing the function, so it cannot satisfy the guard exception.

Fixed in signed commit `b9a5ac1`:

- accepts zero or one literal status argument;
- accepts an optional `--` before one literal status (`return -- 1`);
- rejects extra literal arguments that prevent `return` from returning;
- adds a focused regression test and an explanatory semantic comment.

Fresh verification at `b9a5ac1`:

- Zsh runtime reproduction confirmed the original behavior;
- `gofmt` clean;
- `go test ./...` passed;
- `go vet ./...` passed;
- `go build -buildvcs=false ./...` passed;
- GitHub Actions: Go CI, Docs Generate Check, and Trunk Code Quality all passed.

PR #95, opened 2026-07-26
fix(ci): sync setup-go contract pin

### LINKED ISSUE #91: main is red on build-test: setup-go pin drifted from hardened contract test
## What

`main`'s `build-test` check (go-ci.yml) has been failing since commit 998cea5 (PR #88, "bump actions/setup-go from 6.5.0 to 7.0.0 in the github-actions group", merged 2026-07-22).

## Why

`internal/workflowcontract/wiki_docs_sync_test.go` hardcodes an expected exact pin for `actions/setup-go` in `wiki-docs-sync.yml`:

```
uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
```

Dependabot's grouped `github-actions` update bumped `actions/setup-go` to `7.0.0` across workflow files, including `wiki-docs-sync.yml`, but this hardcoded contract test was not updated to match. Two tests now fail on every run:

- `TestWikiDocsSyncUsesHardenedGitHubAppContract`
- `TestWikiDocsSyncRejectsTriggerIdentityActionAndVerificationMutations/setup_Go_action_mutable_ref`

Both fail with the same root cause: the live workflow file's pin no longer matches the string the test expects.

## Impact

`main` has been red on `build-test` since 2026-07-22. This was found incidentally while verifying it would be safe to mark `build-test` a required status check (z-shell/zsh-lint#90); it is an independent, pre-existing defect and is being reported separately rather than fixed as part of that work.

## Suggested fix

Either update the hardcoded pin in `wiki_docs_sync_test.go` to the new reviewed `actions/setup-go@<v7 sha>` value, or revert `wiki-docs-sync.yml`'s `actions/setup-go` pin to `6.5.0` if this workflow is meant to require manual (non-Dependabot) review for that specific action. The test's name ("hardened contract") suggests the latter may be the intended design, in which case Renovate/Dependabot may need an ignore rule for this specific line rather than a one-time test update.

## References

- Failing run: https://github.com/z-shell/zsh-lint/actions/runs/29967538534
- Introduced by: z-shell/zsh-lint#88
- Found while investigating z-shell/zsh-lint#90

ss-o:
## Summary

Update the hardened wiki-sync workflow-contract expectations to the reviewed `actions/setup-go` v7 commit already used by the live workflow.

This is a one-file prerequisite repair. It does not modify the workflow or include the pending wikidoc promotion.

## Root cause

Dependabot PR #88 updated `actions/setup-go` from v6.5.0 to v7.0.0 in the workflow, but the exact contract expectations retained the previous pinned commit and mutable-ref mutation. Untouched `main` therefore failed two workflow-contract tests.

## Impact

The contract again validates the exact action pin used by the workflow, restoring the full Go test gate on `main` and unblocking the separate exact wikidoc promotion from #94.

## Verification

- reproduced both failures on untouched `origin/main`
- `go test ./internal/workflowcontract -count=1`
- `go test ./...`
- `go vet ./...`
- `GOFLAGS=-buildvcs=false go build ./...`
- `trunk check internal/workflowcontract/wiki_docs_sync_test.go`
- confirmed the resulting contract file exactly matches the green `next` version

Closes #91.

Prerequisite for the exact `main` promotion of #94.

### DIFF
--- a/internal/workflowcontract/wiki_docs_sync_test.go
+++ b/internal/workflowcontract/wiki_docs_sync_test.go
@@ -256,7 +256,7 @@ func wikiDocsSyncContractViolations(t *testing.T, workflow string) []string {
 		{"repository scope", "repositories: wiki"},
 		{"app token action pin", "actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1"},
 		{"checkout action pin", "actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0"},
-		{"setup Go action pin", "actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16"},
+		{"setup Go action pin", "actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e"},
 		{"pull request action pin", "peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1"},
 		{"pull request step ID", "id: sync-pr"},
 		{"generated path scope", "add-paths: community/04_zsh_lint/index.mdx"},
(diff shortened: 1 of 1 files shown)


ss-o:
Post-merge metadata audit: the signed source commit contained no prohibited coauthor trailer, but the GitHub-generated squash message added one to published commit `e340a6cbba21bc053d13a82d451ab51f30916da6`. Protected `main` history was not rewritten. The immediately following #96 merge used an explicit squash subject and one-line body, and published commit `d390a963e5f5c92c0d73054360c1f4d486ffe942` was verified to contain no such trailer. The canonical safeguard is documented in the organization branch-protection runbook.

Issue #263, opened 2026-09-16
Parser gap: alternate for loop does not tolerate a multi-line parenthesized word list


ss-o:
## Summary

The alternate-form `for name ( words ) { ... }` sublist adapter only masks the `(`, `)`, `{`, and `}` positions of the construct. When the parenthesized word list spans more than one line, the synthesized `for name in words; do ... done` retry keeps every original byte between the parens verbatim, including the bare newlines. A bare newline directly after `in`, or between two words in the list, is not valid in that `do`-form even in native Zsh, so the retry fails with an unrelated-looking parse error that then gets rebased back onto the `for` keyword itself, masking the real gap.

## Minimized reproduction

| Source | `zsh -f -n` | Result |
| --- | --- | --- |
| `for x (`, `  a b c) {`, `  print $x`, `}` (four lines, list wraps after the opening paren) | ok | `1:1: `for foo [in words]` must be followed by `do`` |
| `for x (a b`, `) {`, `  print $x`, `}` (list wraps before the closing paren) | ok | same family |
| single-line `for x ( a b c ) { print $x }` | ok | ok |

## Mechanism

`scanForEdits` (`internal/parse/multi_name_for.go`) recognizes the `for name ( words ) {` sugar form and emits four edits: replace `(` with ` in `, replace the byte before `)` with `; `, replace `{` with `do\n`, and replace the byte before `}` with `\ndone\n`. It does not touch any byte strictly between the parens, so a literal `\n` in the word list survives into the synthesized `for name in words; do` text unchanged.

Native Zsh accepts a `\`-continued newline inside that position (`for x in \`, newline, `a b c; do`) but rejects a bare one (confirmed with `zsh -f -n` on both forms). The synthesized retry hits exactly the bare form, so `parse(transformed, name)` fails, and the failure is a second, different-shaped parser error (`` `for foo [in words]` must be followed by `do` ``) that is not itself gated by `scanForEdits`'s own recognition logic. `rebaseForError` (added in #255) correctly maps its position back into the original source, but the position it maps to is the `for` keyword, not the actual newline that broke the retry, so the report reads as if the whole construct failed to parse rather than naming the specific newline placement.

## Classification

[Alternate Forms For Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Alternate-Forms-For-Complex-Commands), the same `for name ( words ) { ... }` sugar as #13, #241, and #255.

## Corpus evidence

[...]

PR #264, opened 2026-09-16
fix(parse): rebase alternate-for retry errors and keep later anonymous-arg blockers

### LINKED ISSUE #255: Parser gap: an alternate for loop breaks a later anonymous function call with arguments inside a function
## Summary

An alternate-form `for name ( words ) { ... }` anywhere earlier in a file makes a later anonymous function called with arguments inside a named function fail, and the reported position points outside the file. Two adapter behaviors combine: `parseMultiNameFor` returns a retry error with the position from its transformed source, which has more lines than the original, and the anonymous-invocation fallback in `parseAnonymousFunctionArgs` masks the rest of the file including the enclosing function's `}`, so it cannot recover once the seed position is wrong.

## Minimized reproduction

Results are from the gitlink base `a202693` and are unchanged on the #235 branch.

| Source | `zsh -f -n` | Result |
| --- | --- | --- |
| `for x ( a ) {`, `}`, `f() {`, `  () { x; } y`, `}` (five lines) | ok | `7:13: statements must be separated by &, ; or a newline` in a five-line file |
| the same without the `for` loop | ok | ok |
| the same with the anonymous call at top level instead of inside `f` | ok | ok |
| the same without the argument `y` | ok | ok |
| `for p ( a b ) { x=1 }`, `f() { x; } y` (two lines) | rejected: `parse error near y` | `5:12: statements must be separated by &, ; or a newline` in a two-line file |

The last row is invalid Zsh and must keep failing, but its position shows the drift on its own: line 5 does not exist.

## Mechanism

`parseMultiNameFor` rewrites `for x ( a ) {` `}` into `for x  in  a ;  do` followed by blank lines and `done`, which adds lines. It rebases node positions on success, but on a retry failure it returns the parser error unchanged (`internal/parse/multi_name_for.go`, the `parse(transformed, name)` error path), so the position `7:13` belongs to the transformed text. The alternate-if adapter already handles the same situation with `rebaseAlternateIfError`.

`parseAnonymousFunctionArgs` then receives that error. `anonymousFunctionInvocationCandidate` clamps the out-of-range offset to the last byte and finds no `}` there. `fallbackAnonymousFunctionInvocationCandidate` finds the `} y` pair, but `prefixEndsWithAnonymousFunction` blanks every byte after the anonymous `}`, which removes the closing `}` of `f()`, so the prefix never parses and the fallback rejects the candidate. The fallback therefore only works for an anonymous call at the top level or at the end of a file, which is why the second and third rows pass.

## Classification

[Alternate Forms For Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Alternate-Forms-For-Complex-Commands) for the loop and [Anonymous Functions](https://zsh.sourceforge.io/Doc/Release/Functions.html#Anonymous-Functions) for the call with arguments. Both parse on their own; this is a composition defect between two existing adapters.

## Corpus evidence

`z-shell/zi` (`bcb20c1`) `zi.zsh`: `.zi-run-plugin-standard-unload-callback` starts at line 1342 and calls `() { eval "$callback"; } "$@"` at line 1347; the file has several alternate `for ( ) { }` loops before it, the last at line 1255. The survey reports `1349:23`, which is not the construct's position. This is the first blocker in the file after the #235 fix.

## Acceptance criteria

- A `gap-<id>-alternate-for-then-anonymous-args-in-function.zsh` fixture fails on the public parse path before the fix and passes `zsh -f -n`.
- `parseMultiNameFor` rebases a retry error position into the original source the way `rebaseAlternateIfError` does, and a test asserts the position for the invalid two-line row above.
- The anonymous-invocation fallback proves the candidate without discarding the enclosing function, or the corrected seed position makes the fallback unnecessary for this shape.
- `zi.zsh` parses past line 1347.

## References

- #241: the alternate `for (( ... ))` sublist form, the same adapter family
- #206: the umbrella report for alternate-form complex commands

ss-o:
## Summary

An alternate `for name ( words ) { ... }` loop earlier in a file could break a later anonymous function call with arguments inside a named function, with the reported error pointing past the end of the file.

Three mechanisms combined to produce this.

`parseMultiNameFor`'s retry returned a raw error from its transformed, line-shifted buffer on failure instead of rebasing the position into the original source, the way `parseAlternateIfBrace` already does. This change adds `rebaseForError` and `rebaseForPos`, mirroring that existing precedent, and moves line-start computation ahead of the parse call so both the success and failure paths can use it.

`parseAnonymousFunctionArgs`'s retry loop discarded all progress and reverted to the very first error once a candidate had been masked and the whole-file retry still failed for an unrelated reason. It now returns that retry's own error instead, so a later, genuinely different parser gap is reported rather than the resolved candidate's stale position. This third mechanism is not one of the two the issue names, but was required to satisfy acceptance criterion 4 against the real corpus file.

Closes #255.

## Update: review remediation

Follow-up commits address the Copilot review on this PR.

**Correctness (review comment):** the primary candidate path seeded a search from the current parse error's position but never confirmed the `}` it found actually closes an anonymous `FuncDecl` before masking the words after it. An ordinary brace group immediately followed by a stray word (`{ :; } argument`) matches the same shape and would have been masked too, hiding the genuine syntax error it names. `parseAnonymousFunctionArgs` now validates every accepted candidate with `prefixEndsWithAnonymousFunction` before trusting a masked retry's error, on the rarer error-return path only, and falls back to the true, unmasked error when a candidate does not validate.

That validation pass also exposed a performance problem in the whole-file fallback candidate scan, which was already on trunk before this PR: it treats any `}` byte as a candidate, including one closing a `${...}` parameter expansion. On `zi.zsh` (`z-shell/zi`, `origin/next`, `bcb20c1`), the previously pushed commit here was already correct (`1457:3`, matching the acceptance criteria below) but slow: 6.29s. The fallback is never load-bearing for the existing test suite once the primary, error-position-seeded path is trusted, so it is removed outright: a mask can only matter at the position of the current parse error (any site before it already parsed cleanly under the real parser, so it cannot be an unhandled invocation; any site after it hasn't been reached yet, so masking it can't affect the current error). Same file, same `1457:3` result, now at 0.26s.

Removing the fallback broke one shape it happened to cover: an invocation word separated from its `}` only by a `\`-newline continuation. Added `logicalLineStart` to walk the seed position back across such continuations before the backward `}` search, with a dedicated regression test (`TestAnonymousFunctionInvocationAcrossLineContinuation`).

**Round 2 (automated re-review of commit 13735e2):** the re-review flagged two issues in `prefixEndsWithAnonymousFunction`'s own closer-retry loop, one as an active review comment and one as a suppressed comment noted in the same review's summary.

`closingToken` extracted the still-needed closer by regex-matching the parse error text's trailing backtick-quoted token. When that closer is itself a literal backtick (an unclosed legacy `` `cmd` `` command substitution), Go's `%#q` verb, which mvdan/sh formats every token with, falls back to a double-quoted Go string instead of the usual backtick-delimited form, so the regex never matched and a genuine nested candidate was wrongly rejected. `closingToken` now recognizes that rendering directly, with a new regression test (`TestAnonymousFunctionInvocationCandidateInsideBacktickSubstitution`).

The suppressed comment noted that the retry loop's `maxClosers` bound (`for i := 0; i < maxClosers; i++`) appended a closer on its final iteration without ever reparsing with it appended, an off-by-one that would spuriously reject a real anonymous invocation nested exactly at that depth, and asked for coverage beyond depth 12. The off-by-one is a loop-structure defect independent of how the bound is chosen, so the loop is restructured to always reparse after every append; the bound itself stays fixed (32, up from 12) rather than input-derived, because no source-derived count reliably bounds the loop (`closingToken` can match errors that are not genuine incompleteness, and adapter rewrites on a truncated prefix can request closers the source never contained). Added `TestAnonymousFunctionArgsRetryHandlesDeepNesting`, asserting a genuine invocation nested 32 deep still validates and its later blocker is reported.

**Round 3 (automated re-review of commit 8f0c059):** the re-review surfaced a suppressed comment, a previously-missed issue in code unchanged since the prior review, anchored at `anonymous_function_args.go:47`, in `parseAnonymousFunctionArgs`'s retry loop.

When a later, non-first candidate failed validation (`prefixEndsWithAnonymousFunction`) or failed to bind to a real `FuncDecl` (`bindAnonymousInvocations`), both failure paths fell back to `firstErr`, the error captured before the very first candidate was ever masked. For a mix of one genuine and one false candidate, that reports the already-resolved first invocation's stale position instead of the second candidate's own, genuine syntax error. Reproduced with the review's own example (`() { :; } good` then `{ :; } argument` on the next line): native zsh reports the error on line 2; this fix reports `2:8`; commit `8f0c059` reported `1:11`, the already-resolved first line.

Added a `seedErr` field to `anonymousInvocationCandidate`, capturing the error active at the moment each candidate is discovered, before it is masked. Both failure paths now return the specific failing candidate's `seedErr` instead of the unconditional `firstErr`; `bindAnonymousInvocations`'s signature changed to also return the candidate it failed to bind, on failure, so the caller can extract its `seedErr`. Added `TestAnonymousFunctionArgsRetryReportsFailedCandidateOwnError`, using the review's own mixed-candidate example as a testdata fixture.

**Round 4 (automated re-review of commit 5fa4f49):** the re-review surfaced a suppressed comment anchored at `anonymous_function_args.go:162`: the fixed `maxClosers` bound of 32 still misreports errors for a genuine anonymous invocation nested exactly 33 brace bodies deep, one past the bound, so `parseAnonymousFunctionArgs` reports that candidate's stale `seedErr` instead of the later blocker.

The round 2 reasoning still explains why a fixed component exists at all: it is a hard cap against `closingToken` matching an error that is not genuine incompleteness and the retry ping-ponging instead of converging. It was wrong to let that same fixed bound also cap genuine brace nesting: the number of `{` bytes in the prefix not yet matched by a `}` byte (`unclosedBraceCount`) is a safe over-count of how many brace bodies still enclose a candidate, and only ever grows the bound, never shrinks it. `maxClosers` is now `32 + unclosedBraceCount(prefix)`, an additive combination: the fixed base still covers the convergence guard and non-brace closers, and brace nesting is no longer capped at an arbitrary depth.

Nesting inside non-brace constructs past the fixed base (34 nested `if` blocks, for example; 33 still validates because the invocation's own `{` adds 1) remains a false rejection when a later syntax error exists in the same file; deriving those depths, or detecting non-progress across retries, is not done in this change and is filed as z-shell/zsh-lint#267 (found by the automated re-review of commit b205ad5). Added `TestAnonymousFunctionArgsRetryHandlesNestingPastFixedBound` for the 33-deep brace case the review found, and `TestAnonymousFunctionArgsRetryHandlesNonBraceEnclosingConstructs`, which guards a genuine candidate enclosed by non-brace constructs, which the brace count cannot cover.

**Round 5 (automated re-review of commit a9adfbc):** the re-review surfaced an active review comment anchored at `anonymous_function_args.go:202`, on `unclosedBraceCount` (the round 4 addition): `bytes.Count(prefix, "{") - bytes.Count(prefix, "}")` is not a safe upper bound, because a `}` inside a comment or quoted string is not a real closer but still subtracts from the count. With 33 enclosing `f() {` bodies followed by a comment line containing 33 `}` bytes, the net count was 0, so `maxClosers` stayed at the fixed 32, one short of the genuine depth, and the candidate was spuriously rejected.

`unclosedBraceCount` is renamed to `braceOpenerCount` and now counts `{` bytes only, with no subtraction. Every enclosing brace construct still open at `close` was opened by exactly one `{` byte in `prefix`, so that count can never be smaller than the true number still open there: a non-syntactic `{` only inflates the bound, which stays safe, and nothing is ever subtracted, so nothing can deflate it below what genuine nesting needs. `prefixEndsWithAnonymousFunction`'s own doc comment explains why a hand-rolled scanner cannot validate a candidate (quote state desyncs, so exactness is not available); a bound does not need exactness, only a one-directional error, which is why counting `{` alone works for the bound even though validating with it would not. Added `TestAnonymousFunctionArgsRetryHandlesCommentBracesBeforeCandidate`, reproducing the exact shape the review reported.

**Convention (review comment):** moved the inline invalid source in `TestParseMultiNameForRebasesRetryErrorPosition` into `testdata/invalid-255-multi-name-for-retry-later-blocker.txt`, matching the testdata-fixture pattern used elsewhere (for example `TestParseAlternateIfRejectsBraceGluedToDoubleBracket`).

## Acceptance criteria

- `ok-alternate-for-then-anonymous-args-in-function.zsh` fails to parse on the base commit and passes `zsh -f -n` and the fixed parser. Verified against a binary built from the pre-fix base (`13:13`, drifted past the file's own length) and from this branch (parses cleanly).
- `TestParseMultiNameForRebasesRetryErrorPosition` asserts the issue's own invalid two-line row is rejected at the correctly-rebased `2:12`, not a drifted out-of-file position.
- `TestAnonymousFunctionArgsRetryReportsLaterBlocker` proves the retry reports a genuinely later, unrelated syntax error (`4:1`) instead of the resolved candidate's stale one.
- `zi.zsh` (`z-shell/zi`, `origin/next`, `bcb20c1`) advances past its previous blocker. Base binary (zsh-lint at merge-base `6f06274`, before this PR): `1349:23`, mid-body of an unrelated, valid anonymous function. Fixed binary: `1457:3`, a distinct later gap now filed as z-shell/zsh-lint#263. The review-remediation commit reaches the same `1457:3` in ~0.26s, down from 6.29s in the previously pushed commit on this PR; the round 3 remediation commit (`5fa4f49`) reaches the same `1457:3` in ~0.28s, no regression. The round 4 remediation commit (`a9adfbc`) reaches the same `1457:3` in ~0.29s, no regression. The round 5 remediation commit reaches the same `1457:3` in ~0.26s, no regression.

## Verification

- `go test -count=1 ./...`: all packages pass, including the review-remediation regression tests.
- Local reference-corpus gate run (`src`, `zd`, `zunit`, `z-a-meta-plugins`, `zsh-fancy-completions`, `zsh-eza`, 18 files): native `zsh -f -n` accepts all files, `zsh-lint-survey` reports 18/18 ok, `zsh-lint --no-config` reports 0 errors, 0 warnings.

## Related

- z-shell/zsh-lint#263: a new, distinct parser gap in the same adapter family discovered while validating acceptance criterion 4 against the real corpus, filed separately rather than folded into this fix.
- z-shell/zsh-lint#265: an anonymous function invocation on the same line as its enclosing backtick substitution is misread as unterminated, discovered while investigating the round 3 review remediation above.
- z-shell/zsh-lint#266: `$((` command substitution starting with a subshell is parsed as arithmetic expansion, a distinct front-end ambiguity discovered in the same investigation as #265.


ss-o:
Re the suppressed finding on `b205ad5` (`anonymous_function_args.go:166`, the fixed 32 still capping non-brace nesting): confirmed and reproduced. For the 34-nested-`if` shape with a later `)` blocker, native `zsh -f -n` reports line 70, this branch reports the invocation's stale `35:11`; depth 33 reports the correct `68:1`. Deferred rather than fixed here: a sound termination criterion (parser progress or cycle detection, or a derived non-brace opener bound) needs the `Pos` semantics of each mvdan/sh error shape `closingToken` can match to be established first, and that is its own change. Filed as #267 with the repro, the `main` baseline (the same stale error at every depth before this PR, the original #255 defect), both candidate designs, and acceptance criteria. The Round 4 paragraph in the PR body is corrected to the verified threshold (34, not 33) and links the issue.

Issue #234, opened 2026-09-16
Parser gap: support function definitions whose name contains a parameter expansion


ss-o:
## Summary

A function definition whose name contains a parameter expansion fails to parse, in both definition forms. The `name()` form is rejected at the name; the `function name` form is accepted as a plain command, so the body brace dangles.

## Minimized reproduction

All rows are `zsh -f -n` valid.

| Source | zsh-lint `main` (`420246f`) |
| --- | --- |
| `_w_${cur}() { :; }` | `1:1: invalid func name` |
| `_w_$cur() { :; }` | `1:1: invalid func name` |
| `function _w_${cur} { :; }` | ``1:25: `}` can only be used to close a block`` |

## Classification

[Functions](https://zsh.sourceforge.io/Doc/Release/Functions.html): the function name word is subject to expansion when the definition is executed, and the shell grammar accepts any word there. `function_compat.go` (#213) already adapts function names for a different shape; this is a new name class.

## Corpus evidence

`z-shell/F-Sy-H` `F-Sy-H.plugin.zsh:387` (`6a022ef`): `_fsh_widget_${cur_widget}() { :; _fsh_zle_highlight }`, the ZLE widget wrapper loop.

## Acceptance criteria

- A `gap-<id>-func-name-expansion.zsh` fixture covering the three rows fails before the fix.
- The resulting tree is a `FuncDecl` whose `Name` carries the original bytes and positions; rules that read the name see the unexpanded word.
- Native-invalid names remain rejected (add an `invalid-<id>-<slug>.txt` case if the adapter widens acceptance).

## References

- #213: previous function-name adapter

Issue #213, opened 2026-09-15
Parser gap: support function name; { list } definitions


ss-o:
## Summary

`function name; { list }`, a function definition whose body follows a `;` terminator, fails to parse.

Split from #206.

## Minimized reproduction

`zsh -f -n` valid.

| Source | zsh-lint `main` (`eb9815f`) |
| --- | --- |
| `function f; { print hi }` | ``1:1: `foo()` must be followed by a statement`` |
| `function f { print hi }` | parses |

The raw front end (mvdan/sh v3.13.1 and v3.14.1, LangZsh) fails the same way.

## Classification

[Functions](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Complex-Commands): `function word ... [ () ] [ term ] { list }`, where `term` is one or more newlines or `;`. The `;` before the body is explicitly permitted.

## Corpus evidence

None in the surveyed z-shell corpus. Reporter-supplied, and the reporter describes it as rare. Low priority; recorded so the gap is not lost.

## Acceptance criteria

- A `gap-<id>-function-semicolon-body.zsh` fixture fails on the public parse path before the fix.
- The resulting tree is a `FuncDecl` with the brace body and original positions.

## References

- #206: originating report

PR #304, opened 2026-09-20
fix(parse): support multi-name function definitions (#251)

### LINKED ISSUE #251: Parser gap: support multi-name function definitions name1 name2 () { }
## Summary

A function definition that names several functions at once, `name1 name2 () { list }`, fails to parse. Zsh defines every listed name with the same body. The `function name1 name2 { list }` spelling already parses.

## Minimized reproduction

All rows are `zsh -f -n` valid. Results are from `main` at `5fbd8fd`.

| Source | Result |
| --- | --- |
| `a b () { : }` | ``1:5: a command can only contain words and redirects; encountered `(` `` |
| `a b() { : }` | ``1:4: a command can only contain words and redirects; encountered `(` `` |
| `ZTST_prep ZTST_clean () { : }` | ``1:22: a command can only contain words and redirects; encountered `(` `` |
| `function a b { : }` | ok |

## Classification

[Functions](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Functions), zshmisc: `word ... () [ term ] { list }` defines a function for each word; `function word ... [ () ] [ term ] { list }` is the equivalent reserved-word form.

## Corpus evidence

`z-shell/zpmod` (`5fab33e`) `vendor/zsh/Test/ztst.zsh:310`: `ZTST_prep ZTST_clean () {`. The file is upstream Zsh's vendored test harness and stops there once #215 lets it parse past line 96. No z-shell-authored site uses the form.

## Acceptance criteria

- A `gap-<id>-multi-name-function.zsh` fixture fails on the public parse path before the fix and passes `zsh -f -n`.
- Both `a b () { }` and `a b() { }` parse, and the result exposes every declared name so that function-scoped rules see each one (a `FuncDecl` per name, or metadata mirroring `AnonymousInvocations` when the AST cannot carry it).
- `ztst.zsh` parses past line 310.

## References

- #234: the other function-name form still open

ss-o:
Closes #251. Refs #234 (the other open function-name form) and #213 (the previous function-name adapter).

## What

`parseMultiNameFunction` (`internal/parse/multi_name_function.go`) accepts the multi-name function definition `name1 name2 () { list }`, which defines every listed name with the same body (zshmisc, Functions: `word ... () [ term ] { list }`). mvdan/sh v3.14.1 (LangZsh) reads the `()` form with one name only and rejects the second name at `(`. The reserved-word spelling `function name1 name2 { list }` already parsed to a `FuncDecl` whose `Names` holds every name; that is the shape this adapter restores, so `functionDeclarationNames` and every consumer that already reads `Names` see each declared name with no new metadata.

## How

- Gate: the parser's exact error `` a command can only contain words and redirects; encountered `(` `` with `()` at the error offset. In the Zsh dialect a `(` in a command is a glob qualifier unless `)` follows directly, so the text plus the site names one construct. `a b ( ) { }` (native parse error) fails on a different error and never enters.
- Site: the run of blank-separated plain words (identifier bytes plus `-.:/+@,%`, no quotes, expansions, globs or `=`) before `()` on the same logical line, `\`-newline continuations included, back to a command start (a newline, `;`, `&`, `|`, `(`, `{` or a backtick). A reserved word before the run (`then`, `do`, `else`, `elif`, `if`, `while`, `until`, `time`, `repeat count`) is the boundary; a run that opens with `for`, `case`, `select`, `function` or a closer is never a definition; an assignment prefix or `!` before the run keeps the parser error, as does a quoted or expanded name (#234's class), so the adapter never accepts more than a single-name definition would.
- Mask: every name but the last becomes same-width spaces, so the retry sees a single-name definition at unchanged offsets. The retry goes through `parseWithAdapters`; one site per pass, further sites re-enter the chain.
- Restore and verify: the retry's tree must hold a `*syntax.Stmt` whose `Cmd` is a `FuncDecl` with `Parens`, no `RsrvWord`, and `Name` at the last name's bytes; it gets `Name = nil`, `Names` built from the original bytes and positions, and both `FuncDecl.Position` and `Stmt.Position` moved to the first name. Otherwise the parser error is returned.

## Verification

- Fixture loop per `parser-gap-workflow.md`: `gap-251-multi-name-function.zsh` passed `zsh -f -n`, failed the survey at `6:5` and passed `TestMinimizedCorpus` as a gap on `main`; after the adapter it parses, is promoted to `ok-multi-name-function.zsh` and added to `requiredFixtures`. `zsh-lint --no-config` on it exits 0.
- `go build ./... && go vet ./... && go test ./...` pass; gofmt clean; `trunk check` on the changed files reports no issues.
- `multi_name_function_test.go`: 26 valid shapes with names, `Names` positions and ends, declaration and statement positions, block span and printed text asserted; three sites in one file restored; single-name and `function a b ()` shapes unchanged; four `invalid-251-*.txt` sources (`x=1 a b () { }`, `a b ( ) { }`, `a b () { } c`, `a b ()`) rejected with the parser's family and position, each confirmed rejected by `zsh -f -n`; unrecognised sites (assignment prefix, `!`, quoted or expanded name, redirect in the run, `repeat 3 () { } a b`, `select a b ()`) hand the error on without a retry; the mask is asserted byte for byte; a foreign retry tree fails closed; native-valid `coproc a b ()`, `nocorrect a b ()`, `local a b ()` are pinned as still rejected (the parser rejects the same prefixes before a single name; separate gaps if ever needed).
- Originating site, `z-shell/zpmod` `vendor/zsh/Test/ztst.zsh:310` (`5fab33e`): `main` (`ffbd25e`) fails at `310:22`; this branch parses the whole file.
- Discovery survey over the local `zi`, `zpmod`, `zunit` checkouts (152 native-valid files): 7 failed on `main`, 6 on this branch; `ztst.zsh` is the only file whose result changed. Reviewed corpus roots from `corpus-paths.txt` (14 files, local checkouts, not the pinned revisions): 14 ok before and after.

## Findings outside this change, tracked separately

1. `a b () ; { : }` and single-name `a () ; { : }` are native-valid but stay rejected: `findFunctionSemicolonBody` (#213) requires the `function` keyword. Filed as #307.
2. `internal/scope/index.go:19` reads `x.Name.Value` without a nil check; it would panic on `function a b { }` and anonymous functions if any rule opted into `NeedsScope` (none does today). Filed as #306.


ss-o (commented):
Fallback review under ADR-0026: Copilot request not registered on 1844fe487159f5d2c87775b1131ee5b29e647578

Review executed under `.github/skills/code-review/SKILL.md` against `code-review-generic.instructions.md` (z-shell/.github, main) and this repository's `AGENTS.md`, `parser-front-end.instructions.md` and `docs/project/parser-gap-workflow.md`. Scope: the full diff of `1844fe4` against `912dbbf` (main). Component class: Go parser front end (class 2 tool); no Zsh source is changed except the corpus fixture.

## Checks that ran

- `go build ./... && go vet ./... && go test ./...`: pass (11 packages).
- `gofmt -l .`: clean. `trunk check` on the 10 changed files: no issues.
- `zsh -f -n internal/survey/testdata/corpus/ok-multi-name-function.zsh` (zsh 5.9.2): valid. `zsh -f -n` on each `invalid-251-*.txt`: rejected, as the fixture contract requires.
- `go run ./cmd/zsh-lint-survey` on the fixture: FAIL at 6:5 on main, OK on this head. `zsh-lint --no-config` on the fixture: exit 0.
- Hosted CI: 59 of 59 checks green on this head.
- Adapter contract (`parser-gap-workflow.md`): registered once in `adapterChain`; retry through `parseWithAdapters`; gate is one exact parser error plus a `()` site check; the mask keeps every byte offset; no comment byte is masked; the retried tree is verified before it is returned and the parser error is returned otherwise; `adapterSnippets` gained the entry `TestAdapterChainCoversEveryAdapter` requires, so the ordered-pair and all-features composition matrix covers the new adapter.
- Reviewer probes beyond the committed tests, all through `Parse`: restored `Names[1]` position equals the parser's own position for the same byte after non-ASCII text on the line (`1:19..1:20` both), so the byte-column convention is consistent; `\r\n` line ending; nested definition `a b () { g h () { : } }`; three sites across a brace group and an `always` block; `repeat a b () { : }` yields a repeat loop with count `a` whose body defines `b`, which is how native Zsh reads `repeat word sublist`.

## Findings

Security: no input execution, no file or network access, no temporary files. State integrity: the adapter mutates only the tree it just built from the retry, and `File` is constructed after all adapters return. Compatibility: no public API or CLI output change; the new tree shape (`Name == nil`, `Names` set, `Parens`, `RsrvWord == false`) is already produced upstream for `function a b ()` and already handled by `functionDeclarationNames`, `fpath_hygiene.go:133` and `anonymous_function_args.go`. Test coverage: 26 valid shapes with positions, a byte-exact mask assertion, three-site restoration, fail-closed retry, gate isolation, four native-invalid sources and three native-valid prefixes pinned as still rejected.

1. `findMultiNameFunction` treats any `\` directly before a newline as a continuation without checking backslash parity (`\\` followed by a newline is an escaped backslash and a real line break). Traced: after the skip, the word scan meets the remaining backslash, which is not a name byte, so the run can never extend across a real line break; the only continuation that joins names is the single `\`, which is correct. No input distinguishes the two readings, so there is nothing a test could pin. Leave as is unless the scanner is shared later.
2. `a '' b () { : }` and `a "b" () { : }` are native-valid and stay parser errors: the name class is deliberately the set the parser itself accepts for a single name, so this is the #234 name class, not a regression.

Evidence gap: the corpus gate's pinned revisions were not checked out locally; the 14 files under `corpus-paths.txt` were surveyed from the workspace checkouts (14 ok before and after). The hosted `corpus` checks on this head are green.

Issue #259, opened 2026-09-16
Parser gap: support let as the last command before } without a separator


ss-o:
## Summary

A brace block whose last command is `let` with no separator before the closing `}` fails to parse, the same way a declaration builtin does in #231. `letClause` in the front end reads arguments until a stop token, and `}` is not one, so the brace becomes one more `let` argument and the block is left open. Found while probing #231 shapes; `let` is a separate builtin (arithmetic evaluation, `zshbuiltins`), so it is filed on its own.

## Minimized reproduction

All rows are `zsh -f -n` valid.

| Source | zsh-lint `main` (`eebfe8a`) |
| --- | --- |
| `{ let x=1 }` | ``1:1: reached EOF without matching `{` with `}` `` |
| `f() { let i++ }` | ``1:5: reached EOF without matching `{` with `}` `` |
| `{ let x=1; }` | parses |
| `{ (( x = 1 )) }` | parses |

## Classification

[Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Complex-Commands), `{ list }`: the closing brace terminates the list without a preceding separator. Same family as #231, different clause parser (`letClause` instead of `declClause`).

## Corpus evidence

None in the workspace corpus (192 files); `let` before a bare `}` does not occur there. Reporter-driven.

## Acceptance criteria

- The two failing rows parse into a `Block` containing one `LetClause` with the original positions preserved; the separator form is unchanged.
- If #231 lands as a separate adapter, extend its gate to `let` rather than adding a second adapter for the same brace-termination construct.

## References

- #231: the declaration-builtin form of the same termination gap

Issue #238, opened 2026-09-16
Parser gap: support do; with a leading separator in loop bodies


ss-o:
## Summary

A loop body that begins with a separator directly after `do` (`do; list`) fails to parse for `while`, `for`, and `until`. This is family H of `docs/project/2026-08-28-discovery-survey.md`, which was recorded but never filed.

## Minimized reproduction

All rows are `zsh -f -n` valid.

| Source | zsh-lint `main` (`420246f`) |
| --- | --- |
| `while (( $# )); do; shift; done` | ``1:1: `while` statement must end with `done` `` |
| `for x in a b; do; print $x; done` | ``1:1: `for` statement must end with `done` `` |
| `until (( $# )); do; shift; done` | ``1:1: `until` statement must end with `done` `` |
| `while (( $# )); do shift; done` | parses |

## Classification

[Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Complex-Commands): `do list done`, where Zsh accepts a leading separator in `list`.

## Corpus evidence

`z-shell/zpmod` `vendor/zsh/StartupFiles/zshrc:45` (`5fab33e`): `freload() { while (( $# )); do; unfunction $1; autoload -U $1; shift; done }`. This is upstream Zsh's shipped example `zshrc`, vendored; no z-shell-authored file in the surveyed workspace uses the form, so this is the lowest-priority open gap.

## Acceptance criteria

- A `gap-<id>-do-leading-separator.zsh` fixture covering the three failing rows fails before the fix.
- The resulting tree is the ordinary `WhileClause`, `ForClause`, or `UntilClause` with the body statements at their original positions.

## References

- `docs/project/2026-08-28-discovery-survey.md`, family H

Issue #231, opened 2026-09-16
Parser gap: support a declaration builtin as the last command before } without a separator


ss-o:
## Summary

A brace block whose last command is a declaration builtin (`typeset`, `local`, `export`, `readonly`, `declare`; `integer` and `float` are not treated as declaration clauses by the front end and already parse) with no separator before the closing `}` fails to parse. The raw front end consumes `}` as one more argument of the declaration clause. Plain commands and assignments in the same position already parse (`ok-brace-termination.zsh`).

## Minimized reproduction

All rows are `zsh -f -n` valid.

| Source | zsh-lint `main` (`420246f`) |
| --- | --- |
| `{ typeset -g C=1 }` | ``1:1: reached EOF without matching `{` with `}` `` |
| `{ local C=1 }` | same |
| `{ export C=1 }` | same |
| `{ typeset -g C=1 } 2>/dev/null` | same |
| `{ typeset -g C=1; }` | parses |
| `{ x=1 }` | parses |

## Classification

[Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Complex-Commands), `{ list }`: in Zsh the closing brace terminates the list without a preceding separator. This is family B of `docs/project/2026-08-28-discovery-survey.md`, which was recorded but never filed.

## Corpus evidence

`z-shell/zi` `lib/zsh/git-process-output.zsh:11` (`bcb20c1`): `{ typeset -g COLS="$(tput cols)" } 2>/dev/null`. The file survey reports a later error first, so this gap is masked until the brace-form `if` defect at line 53 is fixed.

## Acceptance criteria

- A `gap-<id>-brace-decl-termination.zsh` fixture covering the four failing rows fails before the fix.
- The resulting tree is a `Block` containing one `DeclClause` whose last argument is the assignment, with the redirection attached to the block and original positions preserved.
- The separator form `{ typeset -g C=1; }` is unchanged.

## References

- `internal/survey/testdata/corpus/ok-brace-termination.zsh`: the plain-command and assignment cases, covered since the LangZsh switch (`docs/project/2026-06-12-langzsh-switch.md`)
- `docs/project/2026-08-28-discovery-survey.md`, family B

PR #309, opened 2026-09-20
fix(parse): close a declaration brace block inside a loop, if or case body (#298)

### LINKED ISSUE #298: Parser gap: a declaration builtin before a bare } inside a loop or if body is rejected
## Summary

The declaration-brace adapter (#231, #259, `parseDeclarationBraceClose`) closes `{ local x }` only when the parser reports the block as unclosed at EOF, a `)`, a `;;` or a backtick (`unmatchedBraceClose`, `without matching `{` with `}``). Inside a loop or `if` body the parser reaches `done` or `fi` first and reports that keyword instead, so the gate never fires and the source is rejected.

Found while building the #238 fixture; the row was removed from `ok-do-leading-separator.zsh` (#296) because it fails on `main` with or without a leading separator.

## Minimized reproduction

| Source | zsh-lint `main` (`ffbd25e`) |
| --- | --- |
| `while true; do { local x }; done` | `` 1:29: `done` can only be used to end a loop `` |
| `while true; do { local x }`⏎`done` | `` 2:1: `done` can only be used to end a loop `` |
| `if true; then { local x }; fi` | `` 1:28: `fi` can only be used to end an `if` `` |
| `f() { while true; do { local x }; done }` | `` 1:35: `done` can only be used to end a loop `` |
| `{ local x }; print y` | parses |
| `while true; do { local x; }; done` | parses |

## Classification

Same construct as #231 and #259 ([Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Complex-Commands), `{ list }` with a declaration as the last command and no separator); the gap is the adapter's error gate, not a new language feature.

## Corpus evidence

None in the surveyed z-shell corpus.

## Acceptance criteria

- A `gap-<id>-decl-brace-close-in-body.zsh` fixture covering the four failing rows fails on the public parse path before the fix.
- `parseDeclarationBraceClose` also gates on `` `done` can only be used to end a loop `` and `` `fi` can only be used to end an `if` `` (and the `esac`/`}` equivalents if they reproduce), with the same site scan and `restoreDeclarationBraceClose` verification, so an unrelated stray `done` keeps its error.
- Existing `invalid-231-*.txt` and `invalid-259-*.txt` sources stay rejected.

## References

- #231, #259, #262: the adapter and its current gate
- #238, #296: where the shape was found

ss-o:
Closes #298. Refs #231 and #259 (the adapter and its original gate), #238 and #296 (where the shape was found).

## What

`parseDeclarationBraceClose` (`internal/parse/decl_brace_close.go`) also closes a `{ local x }` block that sits inside a loop, an `if` or a `case` body with a separator after the brace: `while true; do { local x }; done`, `if true; then { local x }; fi`, `case x in (x) { local x }; esac`, and the same shape in condition position (`while { local x }; do`). Native Zsh reads `}` in any position; mvdan/sh v3.14.1 reads the declaration clause up to a stop token, swallows the `}`, stops at the `;`, and then meets the enclosing keyword inside the open block, so it reports `` `done` can only be used to end a loop `` (or `fi`, `esac`, `do`, `then`, `elif`) rather than the unmatched-brace error the adapter gated on.

## How

- Gate: the existing `without matching `{` with `}`` suffix, plus the six exact keyword texts (`keywordInsideBlockErrors`). `` `}` can only be used to close a block `` and `` `;;` can only be used in a case clause `` are not in the set; `TestParseDeclarationBraceCloseLeavesOtherErrorsUntouched` and the `;;` invalid source pin that.
- Site: `findDeclarationBraceClose` now takes `after, before` bounds instead of one seed. The unmatched error lands on the open `{`, so the candidate is the first `}` after it (unchanged behaviour); the keyword error lands past the block, so the candidate is the first `}` before it. The scan itself (quote, comment, heredoc, arithmetic and substitution awareness, whole-word `}` followed by a separator, redirect, closer or end of source) is unchanged.
- Mask, verification and restoration are unchanged: the blank before the `}` becomes `;`, the retry goes through `parseWithAdapters`, `restoreDeclarationBraceClose` must find the block at that brace with the synthetic separator on its last declaration statement, and clears it. A second block before the same keyword is reached by re-entry; a stray keyword with no candidate before it, or one that remains after every candidate is closed, keeps the parser's own error at the keyword's own position, which is where native Zsh reports it.

## Verification

- Fixture loop per `parser-gap-workflow.md`: `gap-298-decl-brace-close-in-body.zsh` (27 rows) passed `zsh -f -n` and failed the survey built from `main` at `7:29` (`decl_brace_close.go` is byte-identical between `ffbd25e` and `7ec3bc1`); after the change it parses and is promoted to `ok-decl-brace-close-in-body.zsh` and added to `requiredFixtures`. `zsh-lint --no-config` on it exits 0.
- `go build ./... && go vet ./... && go test ./...` pass; gofmt clean; `trunk check` on the changed files reports no issues.
- `decl_brace_close_test.go`: `TestParseDeclarationBraceCloseInsideBody`, 23 rows asserting the bare parser's error text, every declaration block's `Rbrace` at the source `}`, the last statement without a separator and ending before the brace, and every literal reading back from the source; the same-line shapes without a separator (`{ local x } done`, `{ local x } fi`) pinned as still going through the unmatched-brace path; four `invalid-298-*.txt` sources (`{ local x }; done` at top level, `done` closing an `if`, two loops with a stray final `done`, `;;` after the brace inside a loop) rejected with the parser's family and position, each confirmed rejected by `zsh -f -n`; five sources with no candidate before the keyword or a different error family handed on without a retry.
- Discovery survey over the local `zi`, `zpmod`, `zunit` checkouts (152 native-valid files): 6 failed before and after, no file's result changed. Reviewed corpus roots from `corpus-paths.txt` (14 files, local checkouts): 14 ok before and after.

## Notes

- #296 removed a `while true; do { local x }; done` row from `ok-do-leading-separator.zsh` because of this gap; once both land, that row can return to that fixture.
- `else` never surfaces as its own error text for this shape (the parser reports the following `fi`), so it is not in the gate; the `else` body rows are covered by the `fi` entry.


ss-o (commented):
Fallback review under ADR-0026: Copilot request not registered on 421b9c2dc78fc53dc8bc8223b76b7fbb574110c9

Review executed under `.github/skills/code-review/SKILL.md` against `code-review-generic.instructions.md` (z-shell/.github, main) and this repository's `AGENTS.md`, `parser-front-end.instructions.md` and `docs/project/parser-gap-workflow.md`. Scope: the full diff of `421b9c2` against `7ec3bc1` (main). Component class: Go parser front end (class 2 tool); the only Zsh source added is the corpus fixture.

## Checks that ran

- `go build ./... && go vet ./... && go test ./...`: pass (11 packages). `gofmt -l .`: clean. `trunk check` on the changed files: no issues.
- `zsh -f -n` on `ok-decl-brace-close-in-body.zsh` (zsh 5.9.2): valid. `zsh -f -n` on each `invalid-298-*.txt`: rejected.
- Survey on the fixture: FAIL at 7:29 with the binary built from `main`, OK on this head. `zsh-lint --no-config` on the fixture: exit 0.
- Hosted CI: 60 of 60 checks green on this head.
- Adapter contract: no new adapter, so no chain or snippet change is needed; the gate is widened by exact error texts only (the shared `can only be used` phrase is not matched, so the `}` and `;;` errors stay out, pinned by `TestParseDeclarationBraceCloseLeavesOtherErrorsUntouched` and the `;;` invalid source); the mask keeps every offset; `restoreDeclarationBraceClose` still verifies the block at the brace before the tree is returned; the parser error is returned otherwise.
- Reviewer probes beyond the committed tests, all through the survey: the shape inside `$( ... )` and inside a legacy backtick substitution; a quoted `"done"` and a `done` heredoc line between the block and the real keyword; a `done` inside a trailing comment; `x='done'` as the declared value; a redirect after the block; two blocks on separate lines; a block followed by `always` in a `for` body. All parse. `if true; then { local x }; fi; done` (native-invalid) keeps `` `done` can only be used to end a loop `` at 1:32, the stray keyword's own position.

## Findings

Security: no input execution, no file or network access. State integrity: the retry mutates only the tree it built, and the direction of the site bound is decided once from the error text before the scan runs. Compatibility: no public API or output change; `findDeclarationBraceClose` is package-private and its one direct test caller is updated. Test coverage: 23 body rows with tree-shape assertions, the same-line no-separator shapes pinned on the unmatched-brace path, four native-invalid sources with position assertions, five no-retry sources.

1. `else` is deliberately absent from `keywordInsideBlockErrors`: the parser never reports it for this shape (the `else` rows report the following `fi`, covered). If a future mvdan/sh release changes that text, the `else` body rows in the fixture will fail loudly.
2. Once #296 lands, the `while true; do { local x }; done` row it removed from `ok-do-leading-separator.zsh` can be restored there; not part of this change.

Evidence gap: the corpus gate's pinned revisions were not checked out locally; the 14 files under `corpus-paths.txt` were surveyed from the workspace checkouts (14 ok before and after). The hosted `corpus` checks on this head are green.

ss-o (commented):
Fallback review under ADR-0026: Copilot request not registered on fc170475c26c36dc91801c0135a457bce8329df9

Follow-up to the review of 421b9c2 on this pull request. The new head is a merge of `origin/main` at `b78f5ac` (#296, #305, #308 and the #310 hotfix) into `bug-298`; the #298 change itself is unchanged. The only conflict was the adjacent line in `requiredFixtures` (`internal/survey/corpus_test.go`): both entries are kept, `ok-decl-brace-close-in-body.zsh` and `ok-do-leading-separator.zsh`, in the list's alphabetical order.

Checks on the merged tree: `go build ./... && go vet ./... && go test -count=1 ./...` pass (11 packages), so the widened declaration-brace gate composes with the `do;` separator, `select` short-form and reworked `repeat` adapters through the ordered-pair matrix; the survey passes `ok-decl-brace-close-in-body.zsh`, `ok-do-leading-separator.zsh`, `ok-select-short-form.zsh` and `ok-repeat-synthesized-closer.zsh`; `zsh-lint --no-config` on the #298 fixture exits 0. Hosted CI is green on this head. No new findings.

Issue #300, opened 2026-09-20
Parser gap: repeat sublist ending in an adapter-synthesized closer is mis-closed and panics at end of file


ss-o:
## Summary

`resolveRepeatLoops` places the loop's `done` at the sublist statement's `End()`. When the sublist ends in a construct another adapter rewrote (`for name ( word ... ) { list }` from the multi-name-for adapter, `if list { list }` from the alternate-if adapter), the closing keyword's rebased position is the byte it was written over, and `End()` adds the keyword's length past that byte. The insertion then lands inside the next line, or past the end of the source, where it panics.

## Minimized reproduction

All rows are `zsh -f -n` valid.

| Source | zsh-lint `main` (`ffbd25e`) |
| --- | --- |
| `repeat 3 for i (a b) { print $i }` then `print after` on the next line | ``1:1: `while` statement must end with `done` `` |
| `repeat 3 if (( 1 )) { print hi }` then `print after` on the next line | same |
| `repeat 3; for i (a b) { print $i }` then `print after` on the next line | same |
| `repeat 3 for i (a b) { print $i }` at end of file | panic: `index out of range [36] with length 35` in `resolveRepeatLoops`, `internal/parse/repeat.go:713` |
| `repeat 3 for i in a b; do print $i; done` then `print after` | parses (no synthesized closer) |

## Classification

[Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Complex-Commands): `repeat word sublist`. Not a grammar gap: a positional defect in adapter composition. `ForClause.End()` and `IfClause.End()` add the closing keyword's length to a position that, after `rebaseForPositions`, names the `}` byte the keyword replaced.

## Corpus evidence

None in the surveyed z-shell corpus. Found while implementing #212, whose adapter declines the same shape (`sublistEndsAt`) instead of mis-closing it.

## Acceptance criteria

- A `gap-<id>-repeat-synthesized-closer.zsh` fixture covering the first two rows fails before the fix.
- The end-of-file row parses on the public path without a panic; a parse error is acceptable only while the fixture is still `gap-`.
- The resulting `WhileClause` body holds the whole `for` or `if` statement, with `done` on the byte after its `}`, and `File.RepeatLoops` records the loop.
- The same end computation lets the #212 adapter accept the shape it currently declines (a follow-up there, not this issue).

## References

- #208: repeat adapter (`resolveRepeatLoops`, `repeatCloserOffset`)
[...]

Issue #212, opened 2026-09-15
Parser gap: support the select name in list; sublist short form


ss-o:
## Summary

The `select name in list; sublist` alternate form fails to parse.

Split from #206.

## Minimized reproduction

`zsh -f -n` valid.

| Source | zsh-lint `main` (`eb9815f`) |
| --- | --- |
| `select o in a b c; break` | ``select foo [in words]` must be followed by `do`` |
| `select o in a b c; do break; done` | parses |

The raw front end (mvdan/sh v3.13.1 and v3.14.1, LangZsh) fails the same way.

## Classification

[Alternate Forms For Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Alternate-Forms-For-Complex-Commands): `select name in word ...; sublist`. The reporter's idiom is `select ...; break` followed by a `case` on the chosen value, which avoids nesting the `case` inside the loop body.

## Corpus evidence

None in the surveyed z-shell corpus. Reporter-supplied.

## Acceptance criteria

- A `gap-<id>-select-short-form.zsh` fixture fails on the public parse path before the fix.
- The resulting tree is a `ForClause` with `Select` set whose body holds the single command, with original positions.

## References

- #206: originating report

PR #310, opened 2026-09-20
fix(parse): end a select sublist from the source (#300)

ss-o:
Refs #300, #212, #305 and #308. Hotfix: `main` at `75a1a8f` does not build.

## What

#305 (`select` short form, #212) closed its sublist with `repeatCloserOffset` and `sublistEndsAt`; #308 (#300) replaced that logic in the `repeat` adapter with `repeatSublistEnd`, which scans the sublist's end back from the source, and removed `repeatCloserOffset`. The two merged minutes apart and `main`'s own Go CI and Corpus Gate runs on `75a1a8f` are red with `internal/parse/select_short_form.go:278: undefined: repeatCloserOffset`.

This ports the `select` adapter onto `repeatSublistEnd`, the same end scan the `repeat` and short `if` adapters use: `sublistStatements` also collects the parent map and the comments the scan needs, `sublistEndsAt` is removed, and `docs/project/parser-gap-workflow.md` describes the shared end scan. One behaviour change follows from the port, in the direction #300 intended: a body whose last token is a closing keyword another adapter synthesized no longer keeps the parser error, because the end now comes from the real last byte. `select o in a b c; if (( 1 )) { break }`, the short-`if` body and the alternate-`for` body parse and are pinned as rows; a trailing comment stays inside the body as it does for `repeat`.

## Provenance

The change was authored in the `hotfix-300` task by the session that landed #308; that session was interrupted before committing (owner process gone, task flagged by `ws doctor`), so the task was recovered under the workspace's recovery rule and the tree was validated, committed and published from this session unchanged.

## Verification

- `go build ./... && go vet ./... && go test -count=1 ./...`: pass (11 packages) on this head; `main` at `75a1a8f` fails to build.
- gofmt clean; `trunk check` on the changed files: no issues.
- `TestParseSelectShortForm`: the previously declined "body closed by a synthesized keyword" row moved from `TestParseSelectShortFormDeclines` to three passing rows; the "background binds to the loop" and "trailing comment" expectations follow the source-scanned end.
- Discovery survey over the local `zi`, `zpmod`, `zunit` checkouts (152 native-valid files): 6 failed, the same set as the last green `main` (`7ec3bc1`). Reviewed corpus roots from `corpus-paths.txt` (14 files, local checkouts): 14 ok.


ss-o (commented):
Fallback review under ADR-0026: Copilot request not registered on 2e9280c8442ef6a09b0cac2cc0afacb7eafeba22

Review executed under `.github/skills/code-review/SKILL.md` against `code-review-generic.instructions.md` (z-shell/.github, main) and this repository's `AGENTS.md`, `parser-front-end.instructions.md` and `docs/project/parser-gap-workflow.md`. Scope: the full diff of the head against `75a1a8f` (main). The change was authored in the interrupted `hotfix-300` session and is reviewed here as a second pass by the session that recovered, validated and published it.

## Checks that ran

- `go build ./... && go vet ./... && go test -count=1 ./...`: pass (11 packages) on this head; `main` at `75a1a8f` fails to build with `select_short_form.go:278: undefined: repeatCloserOffset`.
- gofmt clean; `trunk check` on the three changed files: no issues.
- Hosted CI: 62 of 62 checks green on this head, including Go CI and the Corpus Gate that are red on `main`.
- Discovery survey over the local `zi`, `zpmod`, `zunit` checkouts (152 native-valid files): 6 failed, the same set as the last green `main` (`7ec3bc1`). Reviewed corpus roots from `corpus-paths.txt` (14 files, local checkouts): 14 ok.
- Port correctness: `repeatSublistEnd(src, parents, outer, comments, nil)` matches the signature in `repeat.go:963` and the two call sites there; the `parents` map is built the same way (`stack` top before push, popped on the nil visit) and `comments` are the tree's `*syntax.Comment` nodes, which is what the end scan reads. `sublistStatements` no longer stops the walk at the first match so the map covers the whole tree; the widening loop keeps its semantics (`index = 0; continue` and `break` replace the early returns).

## Findings

Behaviour change, intended by #300 and pinned: a `select` body whose last token is a closing keyword another adapter synthesized (`if (( 1 )) { break }`, the short `if`, the alternate `for`) now parses because the end comes from the real last byte; the former decline row moved to three passing rows. A trailing comment stays inside the body, as it does for `repeat`. The `background binds to the loop` closer moved from 1:29 to 1:28, the byte before the `&`, consistent with the source scan.

1. #305 and #308 were merged within minutes on a shared helper without a re-run against the combined tree. The `Go CI` result on `main` caught it; nothing in this change to do, but the sequencing is worth a line in the learning-capture review.

Evidence gap: the corpus gate's pinned revisions were not checked out locally; the hosted `corpus` checks on this head are green.

Issue #211, opened 2026-09-15
Parser gap: support single-command for, while, and until loop forms


ss-o:
## Summary

The alternate single-command loop forms fail to parse: `for name in list; sublist`, `for name (list) sublist`, `while list sublist`, and `until list sublist`.

Split from #206.

## Minimized reproduction

All rows are `zsh -f -n` valid.

| Source | zsh-lint `main` (`eb9815f`) |
| --- | --- |
| `for t in $a; consume_task $t` | ``for foo [in words]` must be followed by `do`` |
| `for i in {$#a..1}; (( i > 100 )) \|\| a[i]=()` | same |
| `for i (a b) print $i` | ``for foo` must be followed by `in`, `do`, `;`, or a newline`` |
| `for i (a b) { print $i }` | parses |
| `while (( i < 3 )) (( i++ ))` | ``1:19: statements must be separated by &, ; or a newline`` |
| `until (( success )) retry && success=$REPLY` | ``1:21: statements must be separated by &, ; or a newline`` |
| `while (( $# )) { shift }` | parses (#125) |

The raw front end (mvdan/sh v3.13.1 and v3.14.1, LangZsh) fails the same way. Upstream tracks this family as mvdan/sh#1203 (open since 2025-10, no maintainer response).

## Classification

[Alternate Forms For Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Alternate-Forms-For-Complex-Commands): `for name ( word ... ) sublist`, `for name in word ...; sublist` (the second form requires a `;`), `while list sublist`, `until list sublist`; the `while`/`until` forms require a suitably delimited `list`.

## Corpus evidence

None in the surveyed z-shell corpus (the 39 `for x ( ... ) {` sites in `zi.zsh` already parse). Reporter-supplied.

## Acceptance criteria

- A `gap-<id>-loop-short-form.zsh` fixture fails on the public parse path before the fix and covers all four productions.
- Resulting trees are `ForClause` / `WhileClause` (with `Until` set) whose body holds the single command, with original positions.
- Native-invalid undelimited forms such as `while true print x` remain rejected.

## References

- #13: multi-name `for` adapter, which already rewrites `for` headers
- #206: originating report
- mvdan/sh#1203 (tracking reference only)

PR #313, opened 2026-09-20
fix(parse): accept a select header with an empty body (#302)

### LINKED ISSUE #302: Parser gap: support a select header with an empty body
## Summary

A `select` header followed by an empty body fails to parse. Native `par_for` reads one sublist after the header, and `par_sublist` accepts an empty one, so a header at end of file or directly before a closer is valid.

Split from #212, whose adapter requires a statement at the body's first byte.

## Minimized reproduction

| Source | zsh-lint `main` (`ffbd25e`) |
| --- | --- |
| `select o in a b c;` at end of file | ``1:1: `select foo [in words]` must be followed by `do` `` |
| `f() { select o in a b c; }` | ``1:7: `select foo [in words]` must be followed by `do` `` |
| `select o in a b c; do; done` | parses |

`select o in a b c;; break` is native-invalid and must stay rejected.

## Classification

[Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Complex-Commands): `select name [ in word ... term ] sublist` with an empty sublist. The repeat adapter already accepts the same shape for `repeat 2` at end of file (#208).

## Corpus evidence

None in the surveyed z-shell corpus. Split from #212 during implementation; lowest priority of the select follow-ups.

## Acceptance criteria

- A `gap-<id>-select-empty-body.zsh` fixture covering both rows fails on the public parse path before the fix.
- The resulting tree is a `ForClause` with `Select` set and an empty `Do`, positioned on the original bytes.
- An `invalid-<id>-*.txt` fixture keeps `;;` after the header rejected.

## References

- #212: select short form
- #208: repeat adapter, empty-body precedent

ss-o:
Closes #302

## Summary

A `select name [in word ...]` header followed by an empty body is native-valid Zsh: par_for reads one sublist after the header and par_sublist accepts an empty one, so `select o in a b c;` at end of file, `f() { select o in a b c; }`, `( select o in a b c; )`, a header before `done`/`fi`/`esac`, and `if select o in a; then :; fi` all pass `zsh -f -n`. The #212 adapter required a statement at the body's first byte and kept the parser error otherwise.

- `parseSelectShortFormWithParser` now reads an empty body when the byte-preserving probe parses with nothing starting at the body's first byte **and** that byte is the end of the file or a closer (`}`, `)`, `done`, `fi`, `esac`, `elif`, `else`, `then`; `selectEmptyBodyAt`). `do` and `done` are inserted on that byte, so `DoPos` and `DonePos` rebase to it; `verifySelectLoop` requires a select loop there with an empty `Do`.
- Anything else at that byte keeps the parser error. Two shapes drove that guard: `! select o in a b; break` (native-invalid; the probe's `break` statement starts at the `!`, so "nothing at the body's first byte" alone would have accepted it with a wrong tree) and `;;` after the header (native-invalid even inside a case arm).
- Consecutive headers nest as native Zsh nests them (`functions j` on `j() { select o in a; select p in b; }` shows the second loop as the first one's body), so only the inner body is empty.

## Fixtures and tests

- `internal/survey/testdata/corpus/ok-select-empty-body.zsh`: 14 rows, whole file passes `zsh -f -n` 5.9.2; it failed the survey at `8:7` before the change.
- `internal/parse/testdata/invalid-302-*.txt`: `;;` after the header, `; ;;`, `; &`, and an unterminated function body; each fails `zsh -f -n` and is asserted with the parser's error and position.
- `select_empty_body_test.go`: tree-shape rows (positions of `select`, `do`, `done`, statement count, and the statements after the loop), the closer guard table, and a fail-closed case through the injected parser. The two empty-body rows in `TestParseSelectShortFormDeclines` moved to the accepting side.

Out of scope: `while select o in a; do :; done`, which Zsh reads through SHORT_LOOPS (the `do ... done` belongs to the select), is #211's shape.

## Verification

`go build ./... && go vet ./... && go test ./...` pass; `gofmt` and `trunk check` clean on the changed files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ss-o:
Correction to this description: `! select o in a b; break` is **native-valid** Zsh, not native-invalid. `zsh -f -n` exits 1 for any top-level negated command (`! true`) without a parse error, because `-n` still applies the `!` to the not-executed command's status; the exit code misled me. The behaviour the guard in this PR keeps (the adapter declining a negated loop) is therefore a parser gap, not an accepts-invalid fix; it is tracked in #321 together with the other negated adapter forms. The code comments in `select_short_form.go` describe it accurately ("a body the probe could not isolate"); only this description was wrong.

Issue #297, opened 2026-09-20
Parser gap: an if branch that begins with a separator after then or else is rejected


ss-o:
## Summary

A `then`, `else` or `elif ... then` branch that begins with a separator (`then; list`) fails to parse. Same root cause as #238 (`do; list`), different construct: mvdan/sh `followStmts` (v3.14.1, LangZsh) returns an empty list on a `;` directly after the reserved word and then demands the closer, so `then; fi` parses while `then; print x; fi` does not.

Split from the #238 verification; the `do` adapter of #296 deliberately hands these errors on without a retry.

## Minimized reproduction

All rows are `zsh -f -n` valid.

| Source | zsh-lint `main` (`ffbd25e`) |
| --- | --- |
| `if true; then; print x; fi` | `` 1:1: `if` statement must end with `fi` `` |
| `if true; then x; else; y; fi` | `` 1:1: `if` statement must end with `fi` `` |
| `if true; then x; elif false; then; y; fi` | `` 1:1: `if` statement must end with `fi` `` |
| `if true; then`⏎`; print x; fi` | `` 2:1: `;` can only immediately follow a statement `` |
| `if true; then; fi` | parses |

## Classification

[Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Complex-Commands): `if list then list [elif list then list] ... [else list] fi`, where a list is a sequence of zero or more sublists, so a leading separator is an empty sublist.

## Corpus evidence

None in the surveyed z-shell corpus; found while probing the #238 shape.

## Acceptance criteria

- A `gap-<id>-then-leading-separator.zsh` fixture covering the four failing rows fails on the public parse path before the fix.
- The resulting tree is the ordinary `IfClause` with the branch statements at their original positions.
- The adapter follows the `parseDoLeadingSeparator` shape (#296): gate on the two error texts, mask the one `;` with a blank, verify the `ThenPos` or `else` position in the retry's tree. Consider sharing the mask and verify steps with the `do` adapter rather than a second copy.

## References

- #238, #296: the `do` form and its adapter
- mvdan/sh `syntax/parser.go` `followStmts`: `if p.got(semicolon) { ... return nil, nil }` under `LangZsh | LangMirBSDKorn`

PR #328, opened 2026-09-21
refactor(parse): share one command-position walker (#299)

### LINKED ISSUE #299: Share one command-position walker between the repeat, if short form and do separator adapters
## Summary

Two open pull requests each carry a copy of the command-position source walk that `scanRepeatSites` (`internal/parse/repeat.go`) introduced:

- #295 (bug-210) extracts it from `scanRepeatSites` as `scanCommandWords(src, visit)` and uses it for the if short form.
- #296 (bug-238) copies the same body as `scanDoSeparatorWords` in `do_leading_separator.go`, named distinctly so both branches build on their own.

Once both are on `main`, drop the copy in `do_leading_separator.go` and call `scanCommandWords`. The visit signature is identical, so the change is the function name plus deleting the duplicate. #297 (the `then;`/`else;` form) should build on the shared walker from the start.

## Acceptance criteria

- One walker in `internal/parse`; `scanDoSeparatorWords` removed.
- `go test ./internal/parse/` unchanged, including `TestAdapterCompositionAllOrderedPairs`.

## References

- #295, #296, #297

ss-o:
## Summary

`scanDoSeparatorWords` (`internal/parse/do_leading_separator.go`) was a copy of the command-position walk that `scanCommandWords` (`internal/parse/repeat.go`) already provides. The copy was deliberate: #295 (bug-210) extracted the walk from `scanRepeatSites` as `scanCommandWords`, and #296 (bug-238) landed the same body under a distinct name so both branches could build in parallel. Both are on `main` now, so this drops the copy.

`findLeadingSeparatorSite` calls `scanCommandWords`; the duplicate function and its doc comment are deleted. The visit signature is identical, so the change is the call site plus the deletion.

## Verification

- The two bodies were compared before deleting: byte-identical apart from the function name (4493/4494 bytes, the one-byte difference being the name length).
- `go build ./...`, `go vet ./...` clean; `gofmt -l internal/` empty.
- `go test ./...` — every package ok.
- `go test ./internal/parse/ -v` reports **259 passing tests before and after** the change, so the test surface is unchanged, not merely green.
- `TestAdapterCompositionAllOrderedPairs` PASS (the acceptance criterion).
- No residual `scanDoSeparatorWords` reference remains in any `.go` or `.md` file.

## Acceptance criteria

- [x] One walker in `internal/parse`; `scanDoSeparatorWords` removed.
- [x] `go test ./internal/parse/` unchanged, including `TestAdapterCompositionAllOrderedPairs`.

Closes #299


ss-o:
Reviewed as the ADR-0026 fallback reviewer (Copilot rate-limited). **Approve** — the change is what the issue asked for and the equivalence is proven, not assumed.

## What was checked

**The two bodies really were duplicates.** Extracted both functions from `origin/main` and compared with the name normalised away: identical, 4493 vs 4494 bytes, the single byte being `scanCommandWords` vs `scanDoSeparatorWords`. So deleting one and repointing the call site cannot change behavior by construction.

**No orphans left behind.** `scanDoSeparatorWords` has no remaining reference in any `.go` or `.md` file, including the doc comment on `scanCommandWords` in `repeat.go`, which described the copy as temporary and no longer needs amending — it never named the copy. The helpers the deleted body used (`pendingHeredoc`, `consumeHeredocBodies`, `parseHeredocDelimiter`, `isRepeatWordBoundary`, `isRepeatWordEnd`, `isIdentByte`, `repeatCommandPrefixWords`) all keep other non-test callers, so nothing became dead code. `go vet ./...` is clean.

**Behavioral equivalence, measured.** Built the survey CLI twice from this worktree — once with the file restored to `origin/main`, once at this PR's head — and ran both over every fixture and corpus file in the tree (157 files: `internal/survey/testdata`, `internal/parse/testdata`, `examples`). Diagnostics and exit codes are **byte-identical across all 157 files**.

**Test surface unchanged, not merely green.** `go test ./internal/parse/ -v` reports 259 passing tests both before and after, so no test silently stopped running. `TestAdapterCompositionAllOrderedPairs` passes (the issue's acceptance criterion), as do the `do`/`then`/`else` leading-separator suites that exercise `findLeadingSeparatorSite` through the shared walker. Full `go test ./...` ok in every package; `gofmt -l internal/` empty.

## Notes, none blocking

- `isDoEmptyBodyError` (`do_leading_separator.go:115`) now has only a test caller; `parseDoLeadingSeparatorWithParser` reaches the same predicate through `isLeadingSeparatorError`. It is a deliberate named wrapper the focused test asserts against, so it is in scope to keep and out of scope for this PR either way.
- The remaining `hasLoopDo` / `leadingSeparatorSpec` structure is untouched, so the `then`/`else` adapter (#297) that shares the spec is unaffected.

## Acceptance criteria

- [x] One walker in `internal/parse`; `scanDoSeparatorWords` removed.
- [x] `go test ./internal/parse/` unchanged, including `TestAdapterCompositionAllOrderedPairs`.

All 72 required checks are green at head `3465add`, which is the exact commit verified locally.

Issue #287, opened 2026-09-17
Docs: `zsh -f -n` expands top-level simple-command words, so the native gate is not a pure syntax check


ss-o:
## Summary

`docs/project/parser-gap-workflow.md` and `.github/instructions/parser-front-end.instructions.md` treat `zsh -f -n` as the syntax oracle: a file it rejects "is a broken script, not a parser gap", and the native-invalid fixture rule records its decision. For a top-level simple command, `zsh -n` still runs word expansion, so its verdict can come from the substitution stage rather than the lexer, in both directions. Found while remediating the review on #286 (#277).

## Observation

Same line, two placements, zsh 5.9.2 (`zsh -f -n`):

| Source | Result |
| --- | --- |
| `print ${m[(r)a,'x]y']}` | `bad substitution`, exit 1 |
| `f() { print ${m[(r)a,'x]y']}; }` | exit 0 |
| `if true; then print ${m["]"]}; fi` | exit 0 |
| `print ${m[x"[y"]}` | `invalid subscript`, exit 1 |
| `x=${m[(r)a,'x]y']}` | exit 0 |
| `print ${x:?msg}` | exit 0 |
| `print $((1/0))` | exit 0 |

`bad substitution` and `invalid subscript` are `subst.c` and `params.c` messages, not lexer errors; the lexer accepts every row (the function-body placement proves it). Expansion of a top-level simple command runs under `NO_EXEC`, with individual stages gated on the option (`subst.c` calls `arithsubst` and raises the `:?` error only `if (isset(EXECOPT))`, which is why `$((1/0))` and `${x:?msg}` stay silent), so the rejection depends on where the line sits and on which substitution stage the shape reaches.

## Consequences for the documented workflow

- A `gap-*` or `ok-*` fixture written as top-level commands can be rejected by the gate for an expansion reason. The subscript shapes above lex but can never run, so rejecting them is the useful answer, but the doc attributes it to syntax.
- A fixture that wraps the same lines in a function passes the gate although its substitutions are always invalid. `TestMinimizedCorpus` and the repository-wide `zsh -n` gate cannot tell the two apart.
- An adapter probe that compares the parser with `zsh -f -n` must place the row at top level to see the expansion verdict, and inside a function to see the lexer verdict. #286 records the quoted-`]` rows as expansion errors for this reason.

## Proposed change

[...]

Issue #261, opened 2026-09-16
Parser gap: support the for name in words; { list } brace-body sublist form


ss-o:
## Summary

The `for name in word ...; { list }` form fails to parse: after the required `;` or newline, the front end sees the `{` and rejects it with ``for loops with braces are a bash/mksh feature; tried parsing as zsh``. `forClause` in mvdan/sh accepts a `{` after the word list for every dialect but then calls `checkLang(pos, langBashLike|LangMirBSDKorn, "for loops with braces")`, which excludes LangZsh. The `for name (word ...) { list }` form parses because it takes a different path. Found while probing #231 shapes; the single-command sublist of the same production is #211, which reports a different error and so needs a different fix.

## Minimized reproduction

All rows are `zsh -f -n` valid unless marked.

| Source | zsh-lint `main` (`eebfe8a`) |
| --- | --- |
| `for f in a b; { print $f }` | ``1:15: for loops with braces are a bash/mksh feature; tried parsing as zsh`` |
| `for f in a b` newline `{ print $f }` | ``2:1: for loops with braces are a bash/mksh feature; tried parsing as zsh`` |
| `for f (a b) { print $f }` | parses |
| `for f in a b; do print $f; done` | parses |
| `for f in a b { print $f }` (invalid: no separator before `{`) | ``1:1: `for foo [in words]` must be followed by `do` `` (correctly rejected) |

The raw front end (mvdan/sh v3.14.1, LangZsh) fails the same way.

## Classification

[Alternate Forms For Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Alternate-Forms-For-Complex-Commands): `for name in word ...; sublist`, where the sublist is a `{ list }` brace block. The separator before the brace is mandatory; the last row shows the parser already rejects its absence.

## Corpus evidence

None in the workspace corpus (192 files); the `;` and newline forms do not occur there. Probe-driven.

## Acceptance criteria

- Both failing rows parse into a `ForClause` with `Braces` set and the body statements at their original positions.
- `for name in word ... { list }` without a separator stays rejected.
- `for name (word ...) { list }` and the `do ... done` form are unchanged.

## References

- #211: single-command sublist of the same production, different parser error
- #241: the arithmetic `for (( ... )) { list }` form, same `checkLang` rejection
- #206: originating report for the alternate forms

Issue #241, opened 2026-09-16
Parser gap: support the alternate for (( ... )) sublist form


ss-o:
## Summary

The alternate `for (( [expr1] ; [expr2] ; [expr3] )) sublist` form fails to parse, both with a brace body and with a single command. mvdan/sh recognizes the brace body only for its bash and mksh dialects and rejects it under LangZsh. Exposed in `z-shell/zi` `lib/zsh/install.zsh` once #216 let the file parse past line 163. Split from #206 like #211, which covers the `for name`, `while`, and `until` short forms but not the arithmetic header.

## Minimized reproduction

All rows are `zsh -f -n` valid unless marked.

| Source | zsh-lint `bug-216` (`1b31217`) |
| --- | --- |
| `for (( i = 1; i < 3; i++ )) { print $i }` | ``1:29: for loops with braces are a bash/mksh feature; tried parsing as zsh`` |
| `for (( i = 1; i < 3; i++ )) print $i` | ``1:1: `for foo [in words]` must be followed by `do``` |
| `for (( i = 1; i < 3; i++ )); do print $i; done` | parses |
| `for i (1 2) { print $i }` | parses |

The raw front end (mvdan/sh v3.14.1, LangZsh) fails the same way.

## Classification

[Alternate Forms For Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Alternate-Forms-For-Complex-Commands): `for (( [expr1] ; [expr2] ; [expr3] )) sublist`.

## Corpus evidence

`z-shell/zi` (`next`, `bcb20c1`): `lib/zsh/install.zsh:215`, `lib/zsh/autoload.zsh:1565`, `zi.zsh:2700`, `zi.zsh:2701`, `zi.zsh:2760`, `zi.zsh:2764`. All six use the brace body.

## Acceptance criteria

- A `gap-<id>-arith-for-sublist.zsh` fixture fails before the fix and covers the brace-body and single-command forms.
- The resulting tree is a `ForClause` with a `CStyleLoop` and the body statements at their original positions.
- `for (( ... )); do ...; done` behavior is unchanged.

## References

- #206: originating report
- #211: the other short forms from the same manual section
- #13: multi-name `for` adapter, which already rewrites `for` headers

PR #329, opened 2026-09-21
fix(parse): accept an empty loop body after while, until and for (#327)

### LINKED ISSUE #327: Parser gap: support an empty body after a while, until or for header
## Summary

A `while`, `until` or `for` header whose body is the empty sublist native Zsh reads at end of file or before a closer is rejected. `par_while` and `par_for` in `parse.c`, when neither `do` nor `{` follows and `SHORT_LOOPS` is set, read one sublist, and an empty sublist there is valid: `while true print x` is native-valid with `true print x` as the whole condition and no body. The `select` analogue was #302 and #319; the arithmetic `for (( ... ))` form already handles it through `selectEmptyBodyAt` (#241).

Found while verifying #211, whose acceptance criteria called `while true print x` native-invalid; it is not.

## Minimized reproduction

Every row has empty `zsh -f -n` stderr (zsh 5.9.2). Measured on `c6e4ca6` with #326 merged locally.

| Source | zsh-lint |
| --- | --- |
| `while true print x` | `` 1:1: `while <cond>` must be followed by `do` `` |
| `while true` | same |
| `while (( 1 ))` | same |
| `while [[ -n $x ]]` | same |
| `until (( 1 ))` | `` 1:1: `until <cond>` must be followed by `do` `` |
| `for i in a b;` | `` 1:1: `for foo [in words]` must be followed by `do` `` |
| `for i in a b` | same |
| `for i in a b print $i` (the words join the list) | same |
| `for i;` | same |
| `for i (a b)` | `` 1:1: `for foo` must be followed by `in`, `do`, `;`, or a newline `` |
| `f() { while true }` | `` 1:18: `}` can only be used to close a block `` |
| `{ for i in a b; }` | `` 1:3: `for foo [in words]` must be followed by `do` `` |
| `if true; then for i in a b; fi` | `` 1:15: `for foo [in words]` must be followed by `do` `` |
| `for (( ; ; ))` | parses (#241) |
| `select o in a;` | parses (#302) |

## Classification

[Alternate Forms For Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Alternate-Forms-For-Complex-Commands): `while list sublist`, `until list sublist`, `for name in word ...; sublist`, `for name ( word ... ) sublist`, with the sublist empty. Note that an undelimited `while` condition followed by a word on the same line (`while true print x`) is this production, not the delimited short form of #211: Zsh has no way to end the condition list before the newline.

## Corpus evidence

None in the surveyed z-shell corpus. Found by probing; a real script hits it only when a loop is written with no body, so the value is in the oracle agreement (the `if` and `select` adapters already map an empty body), not in corpus frequency.

## Acceptance criteria

- An `ok-loop-empty-body.zsh` fixture covering the `while`, `until`, `for ... in`, `for ... (...)` and positional `for` rows parses; each tree is the `WhileClause` or `ForClause` with an empty `Do`, `do` and `done` on the byte native Zsh would run the empty sublist at, and every other position original.
- `while true print x` yields a `WhileClause` whose `Cond` is the single command `true print x` and whose `Do` is empty.
- The `if` short form (#210) and the delimited `while`/`until` short form (#211) are unchanged: a delimited condition followed by a word on the same line still yields that word's statement as the body.

## References

- #211: delimited single-command loop forms (merged in #325)
- #302, #319: the `select` empty body
- #241: the arithmetic `for` empty body (#326)
- #206: originating report

ss-o:
Closes #327.

Native Zsh's `par_while` and `par_for` read one sublist when neither `do` nor `{` follows the header and `SHORT_LOOPS` is set, and an **empty** sublist there is valid. The parser (mvdan/sh through v3.14.1) instead requires `do`, so every row below passed `zsh -f -n` and failed `zsh-lint`.

## What the parser reports, and why there are two gates

An empty body produces two different errors depending on what follows the header:

- at the keyword — `` `while <cond>` must be followed by `do` `` (e.g. `while true` at EOF)
- at the **closer**, because the parser takes the closer itself as the body and rejects it where it stands — `` `}` can only be used to close a block `` (e.g. `f() { while true }`)

Both are gated. The second gate is a position gate, which the parser front-end contract allows for a construct whose error depends on its body's first token, provided the retry's tree is verified to hold the expected node at each site. `verifyWhileEmptyBodies` does that: every site must come back as a `WhileClause` at its keyword, of the right kind, with `len(Do) == 0` and `DoPos == DonePos` on the body's byte, or the original parser error stands.

## `for` versus `while`

The `for` family already gated correctly and only bailed where the body would start, so it now maps the empty body through `selectEmptyBodyAt` — the same helper the merged `select` (#302/#319) and arithmetic `for` (#241) adapters already use. `forDoneText` places `done` on the body's own byte instead of a new line, since an empty body puts `do` and `done` at the same offset.

`while`/`until` needs a new adapter because **a condition's extent cannot be read from the bytes alone**: in `while true print x` the condition is the single command `true print x`, not a body. So the adapter parses a byte-preserving probe with every candidate keyword blanked (`while` and `until` are both five bytes, so all offsets are preserved), letting the parser itself report where each condition ends, then inserts `do`/`done` at the confirmed sites through one source map and rebases.

## One pass, not one site per pass

My first working version resolved one site per chain re-entry. It was correct but cost a parse pair per site — exponential in their number:

| sites | one-site-per-pass | single pass |
|---|---|---|
| 2 | 4 ms | 4 ms |
| 10 | 98 ms | 3 ms |
| 15 | no result (timeout) | 3 ms |
| 20 | — | 4 ms |

All sites now resolve in one pass.

## Two exclusions the composition gate found, not me

Headers that already parse must keep their keyword. `TestAdapterCompositionAllOrderedPairs` failed on `while (( $# )); do; shift; done` because blanking every `while` orphaned the `do` of ordinary do-form loops; `scanWhileDoForms` now pairs each keyword with the `do` it owns. The delimited short form of #211 is excluded too, since blanking its keyword leaves the condition and its sublist adjacent with no separator between them. Both exclusions exist because the gate caught them — worth noting as evidence the gate is doing its job.

## Verification

- **Dual oracle, files not `zsh -c`.** Every row was checked against zsh 5.9.2 as a file; `zsh -c` lacks a trailing newline, which changes the verdict (the trap from #287). This caught two rows I had written wrong — `p() { for i in a b }` and `s() { for i in a b print $i }` are natively **invalid**, because an unterminated `in` word list swallows the `}`. They were dropped rather than asserted.
- **23-row corpus fixture** `ok-loop-empty-body.zsh` parses and passes `zsh -f -n`.
- **5 native-invalid fixtures** (`invalid-327-*.txt`) stay rejected, each verified `zsh -f -n` = 1.
- **No behavior change elsewhere:** survey output over the 54 pre-existing corpus and fixture files is **byte identical** between base and this head.
- **Mutation-checked:** removing `parseWhileEmptyBody` from the chain makes the new tests fail, so they have teeth rather than passing vacuously.
- `go build`, `go vet`, `gofmt -l`, and `go test ./...` all clean, including all 552 ordered adapter-composition pairs.

## One inverted assertion, flagged for review

`TestParseWhileShortFormRejectsUndelimited` asserted that `while true print x` stays rejected and described it as "out of scope (empty-body case)". That row is precisely this gap, so it becomes `TestParseWhileShortFormUndelimitedIsEmptyBody` and asserts the tree shape (condition is one `CallExpr` of 3 words, body empty). This is the only place an existing assertion was inverted; please check it deliberately.

## Out of scope, filed separately

Three neighbouring shapes are native-valid and still rejected. Each is a distinct defect, not part of this gap, and none is silently swept in:

- a **newline body** — `while cond` then a command on the next line, which native Zsh runs as the loop's body (verified by execution: it loops)
- a trailing **`&&`/`||`** operator after the condition
- a redundant **`;`** after the condition (`while true; ;`)

Also unchanged and still failing: #261 (`for i in a b; { print $i }`), confirmed pre-existing by comparing survey binaries before and after this change.


ss-o:
Reviewed as the ADR-0026 fallback reviewer (Copilot is rate-limited). GitHub refuses `--approve` on one's own PR, so this comment is the review; the approving verdict below is mine as maintainer, recorded here rather than as a formal review state.

**Verdict: approve**, with one finding that is *not* a blocker on this PR and is filed as #333.

## What I checked, and how

**CI is real and on the right commit.** 73 checks, all `SUCCESS` — none skipped or neutral (`group_by(.conclusion)` = `{"SUCCESS": 73}`). `headRefOid` is `1e7fdfd`, byte-identical to the local commit, so CI tested this tree and not an approximation of it.

**The tests have teeth.** I removed `parseWhileEmptyBody` from the adapter chain and re-ran: the new tests fail (`positions.zsh:1:21: \`}\` can only be used to close a block`). Restored, they pass. A test file that cannot fail is worse than none, so this was worth proving rather than assuming.

**No behavioural change outside the gap.** Survey output over the 54 pre-existing corpus and fixture files is byte-identical between `ba9f3ca` and this head.

**No dead code.** Every helper the PR adds or touches has live callers (`selectOperatorAt` 10, `negationBefore` 6, `scanWhileDoForms` 7, `verifyWhileEmptyBodies` 3, `forDoneText` 3, and so on). Nothing was orphaned by the `for` refactor.

**Concurrency and stability.** `go test -race -count=3 ./internal/parse/` is clean, so neither the probe nor the source-map rebase carries hidden shared state.

**The performance claim is measured, not asserted.** Flat ~4 ms from 2 to 20 sites. The earlier one-site-per-pass design was 98 ms at ten sites and produced no result at fifteen; that design was discarded, not shipped.

## Adversarial probing

I ran ten pathological inputs against `zsh -f -n` and the linter rather than only the happy path: nested empty bodies, an empty body inside an empty body, a quoted `while`, `while` as a bare argument, a negated empty body, `case` terminators, deep nesting, an empty file, and a keyword at EOF with no trailing newline. All agree with native Zsh.

Two disagreed. I chased both to a base build rather than trusting the diff:

- `case x in y) while true;; esac` — pre-existing, identical at base and head.
- **CRLF input** — `f() { while true }` with `\r\n` is rejected by base and accepted here, which looks like a regression introduced by this PR. It is not. A CRLF file with **no loop at all** (`f() { print x }`) is already false-accepted at `ba9f3ca`, as is a CRLF `do`-form loop. The parser treats a trailing `\r` as whitespace everywhere. The third row only differed because it previously failed for the unrelated #327 reason, so the CRLF defect was **masked** behind that error and fixing the empty body stopped hiding it.

  This is a false *accept*, the worst direction for a linter, so it should not sit unrecorded: filed as **#333** with the three-row base-versus-head table that establishes it is pre-existing.

## On the substance

Deriving the condition's extent from a blanked probe is the right call, and the PR explains why it must be: `while true print x` has the single command `true print x` as its condition, so no byte-level scan can find the boundary. Blanking is safe precisely because `while` and `until` are both five bytes and every offset is preserved.

The two exclusions are the strongest part of the change, and they came from the composition gate failing rather than from foresight — `while (( $# )); do; shift; done` had its `do` orphaned until `scanWhileDoForms` paired each keyword with the `do` it owns. That gate earned its cost here.

`verifyWhileEmptyBodies` is what makes the closer-position gate acceptable under the parser front-end contract: every site must come back as a `WhileClause` of the right kind with `len(Do) == 0` and `DoPos == DonePos` on the body's byte, or the original error stands. The synthetic separator is cleared, so no position points into inserted text — asserted directly against original offsets in `TestWhileEmptyBodyPositions`.

Dropping the two fixture rows that turned out to be natively *invalid* (`p() { for i in a b }`) was the right call: an unterminated `in` word list swallows the `}`. Asserting them would have baked a wrong belief into the suite.

One deliberate check: `TestParseWhileShortFormRejectsUndelimited` is the single inverted assertion in this PR. Its old claim that `while true print x` is "out of scope" was simply wrong once #327 is fixed, and the replacement asserts the tree shape rather than merely flipping the expectation. Correct.

## Not blocking

- The `while`/`until` adapter and the `for` change could have been separate commits. One commit is defensible since they close a single issue and share `selectEmptyBodyAt`, but a bisect landing here gets both at once.
- #330, #331 and #332 record the neighbouring native-valid shapes this PR deliberately leaves rejected, so the remaining gap is documented rather than silently narrowed.

PR #335, opened 2026-09-21
perf(ci): check every Zsh file in one job instead of one job per file (#334)

### LINKED ISSUE #334: ci: Zsh Syntax Check spends ~949s of runner time to do 0.3s of work
The Zsh Syntax Check workflow is the slowest in the repository while performing almost no work, and its cost grows with every `.zsh` file added.

## Measured cost

`zsh-n.yml` fans out one matrix leg per discovered `.zsh` file. Per leg, from the job timings of run 35553148329:

| step | duration |
|---|---|
| Set up job | 1s |
| Check out code | 1s |
| **`apt update && apt-get install zsh`** | **10s** |
| `zsh -n` on one file | 0s |
| `zcompile` on one file | 0s |
| teardown | 2s |

At 58 files that is **58 legs, ~949s (15.8 min) of runner time**, and it is the critical path of the whole PR check set at **94s**. Runner setup is the entire cost; the syntax check itself is unmeasurable per file.

For comparison, checking all 58 files locally in one process per file:

- `zsh -n`: **172ms**
- `zcompile`: **166ms**

## Why it matters now rather than later

The number of `.zsh` files is the multiplier, and it is growing quickly:

| | 2026-05 | 06 | 07 | 08 | 09-15 | 09-21 |
|---|---|---|---|---|---|---|
| `.zsh` files | 4 | 4 | 13 | 23 | 23 | **58** |

23 to 58 in the last seven days. Each new fixture adds a VM boot and an apt install to every PR.

## What a fix must preserve

- **Every file checked, every failure reported.** The matrix sets `fail-fast: false`; a contributor fixing syntax wants the whole list, not one error per push.
- **`zcompile` output verified.** The per-leg step runs `ls -al` on the `.zwc`, so a silent no-op cannot pass.
- **The `Zsh Syntax Check Complete` context name.** That is what the repository ruleset requires. The per-leg `zsh-n (<path>)` contexts are *not* required and cannot be, since those names change whenever a file is added, renamed or removed — which is why the collapsing job exists.
- **The unfiltered `pull_request` trigger**, for the reason already documented in the workflow: a path-filtered required check never reports on a PR touching no Zsh and leaves it permanently pending.

## Proposed change

Replace the matrix with a single job that iterates the discovered files, collecting failures and emitting `::error file=` annotations so they appear on the file in the PR diff. Fail if zero files are discovered, so a mistake in the `find` expression cannot pass vacuously. Install Zsh through the org `setup-zsh` action already used by `corpus-gate.yml`.

Expected result: ~949s of runner time to ~15s, and a critical path of 94s to roughly 15s, with the saving growing as files are added.

ss-o:
Closes #334.

The Zsh Syntax Check fanned out one matrix leg per `.zsh` file. Each leg spent ~16s on job setup, checkout and `apt-get install zsh` to run `zsh -n` and `zcompile` on **one** file, which takes no measurable time. At 58 files: **~949s of runner time and a 94s critical path** — the slowest workflow in the repository.

Runner setup was the entire cost, and it scaled with the file count: **23 files a week ago, 58 now.**

## Verified locally before pushing

I extracted both `run:` blocks and executed them against this worktree rather than testing in CI:

| case | result |
|---|---|
| `zsh -n`, clean tree | `Checked 58 file(s), 0 failed.` rc=0, **172ms** |
| `zcompile`, clean tree | `Compiled 58 file(s), 0 failed.` rc=0, **166ms** |
| two deliberately broken files injected | **both** reported, `Checked 60 file(s), 2 failed.` rc=1 |
| empty tree (vacuous-pass guard) | `No Zsh files were discovered` rc=1 |

The failure case is the one worth noting: it reports *every* failure, not just the first, and rc=1 proves the collapsed job can still fail. A single job that silently passes would be worse than a slow matrix.

## Behaviour deliberately preserved

- **Every file checked, every failure reported** — what `fail-fast: false` gave. Failures now carry `::error file=` annotations, so they land on the file in the PR diff instead of inside one leg's log. This is strictly better than before.
- **`zcompile` output verified** — each compile must produce a non-empty `.zwc`, as the per-leg `ls -al` did, so a silent no-op cannot pass.
- **`Zsh Syntax Check Complete`** — the job takes this name because it is the context the ruleset requires. I confirmed against the live ruleset (`required_status_checks`) that the per-leg `zsh-n (<path>)` contexts are **not** required, so no ruleset change is needed and no merge can break on the renamed contexts.
- **The unfiltered `pull_request` trigger** and its explanatory comment are untouched.

## Added, not previously present

Both steps now fail when zero files are discovered. Without that, a mistake in the `find` expression would report success having checked nothing — the failure mode a collapsed job makes easy to miss.

## Removed

The separate `zsh-n-complete` collapsing job. It existed only because matrix leg names are unstable in a ruleset's required-check list; with one job, that job *is* the check. No references to `zsh-matrix` or `zsh-n-complete` remain anywhere in the repository.

Zsh installation moves to the org `setup-zsh` action already used by `corpus-gate.yml`, so the install is described in one place.

## Expected effect

~949s of runner time to ~15s, critical path 94s to roughly 15s, with the saving growing as files are added. The proof will be this PR's own run — it triggers on `pull_request`.


ss-o:
Superseded by #336. Identical content; this branch name (`ci-zsh-n-collapse`) cannot pass Policy / Validate Branch Name, which requires a slash separator (`ci/<slug>`), and GitHub does not allow retargeting a PR's head branch. Reopened from `ci/zsh-n-collapse`.

PR #336, opened 2026-09-21
perf(ci): check every Zsh file in one job instead of one job per file (#334)

### LINKED ISSUE #334: ci: Zsh Syntax Check spends ~949s of runner time to do 0.3s of work

ss-o:
Closes #334.

The Zsh Syntax Check fanned out one matrix leg per `.zsh` file. Each leg spent ~16s on job setup, checkout and `apt-get install zsh` to run `zsh -n` and `zcompile` on **one** file, which takes no measurable time. At 58 files: **~949s of runner time and a 94s critical path** — the slowest workflow in the repository.

Runner setup was the entire cost, and it scaled with the file count: **23 files a week ago, 58 now.**

## Verified locally before pushing

I extracted both `run:` blocks and executed them against this worktree rather than testing in CI:

| case | result |
|---|---|
| `zsh -n`, clean tree | `Checked 58 file(s), 0 failed.` rc=0, **172ms** |
| `zcompile`, clean tree | `Compiled 58 file(s), 0 failed.` rc=0, **166ms** |
| two deliberately broken files injected | **both** reported, `Checked 60 file(s), 2 failed.` rc=1 |
| empty tree (vacuous-pass guard) | `No Zsh files were discovered` rc=1 |

The failure case is the one worth noting: it reports *every* failure, not just the first, and rc=1 proves the collapsed job can still fail. A single job that silently passes would be worse than a slow matrix.

## Behaviour deliberately preserved

- **Every file checked, every failure reported** — what `fail-fast: false` gave. Failures now carry `::error file=` annotations, so they land on the file in the PR diff instead of inside one leg's log. This is strictly better than before.
- **`zcompile` output verified** — each compile must produce a non-empty `.zwc`, as the per-leg `ls -al` did, so a silent no-op cannot pass.
- **`Zsh Syntax Check Complete`** — the job takes this name because it is the context the ruleset requires. I confirmed against the live ruleset (`required_status_checks`) that the per-leg `zsh-n (<path>)` contexts are **not** required, so no ruleset change is needed and no merge can break on the renamed contexts.
- **The unfiltered `pull_request` trigger** and its explanatory comment are untouched.

## Added, not previously present

Both steps now fail when zero files are discovered. Without that, a mistake in the `find` expression would report success having checked nothing — the failure mode a collapsed job makes easy to miss.

## Removed

The separate `zsh-n-complete` collapsing job. It existed only because matrix leg names are unstable in a ruleset's required-check list; with one job, that job *is* the check. No references to `zsh-matrix` or `zsh-n-complete` remain anywhere in the repository.

Zsh installation moves to the org `setup-zsh` action already used by `corpus-gate.yml`, so the install is described in one place.

## Expected effect

~949s of runner time to ~15s, critical path 94s to roughly 15s, with the saving growing as files are added. The proof will be this PR's own run — it triggers on `pull_request`.

---

Replaces #335, which is identical in content but was opened from the branch `ci-zsh-n-collapse`. The org `Policy / Validate Branch Name` check requires a slash separator — `ci/<slug>` or `bug-<issue>` — so that branch could never pass, and GitHub does not allow retargeting an existing pull request's head branch. Same two commits, plus the merge of `main` already present on the old branch.


ss-o:
## Review

Self-review under the ADR-0026 fallback, since Copilot is rate-limited. GitHub refuses an approving review on one's own pull request, so this is posted as a comment; treat it as the review of record.

### Verification performed

Measured on this PR's own CI run (`35563877259`), which is the change verifying itself:

| | before | after |
|---|---|---|
| jobs | 58 | **1** |
| runner time | 949s (15.8 min) | **17s** |
| critical path | 94s | **~17s** |

That is a 56x reduction in runner time, and it grows with the file count: the old design paid ~16s of setup per `.zsh` file to run ~0s of work, so the saving scales linearly with every file added.

Correctness of the gate, run end-to-end against the real script rather than asserted from workflow text:

- clean tree: `Checked 58 file(s) with zsh -n, 0 failed.` rc=0
- one broken file planted: `Checked 59 file(s) with zsh -n, 1 failed.` rc=1, with an `::error file=` annotation that renders inline on the diff

Full Go suite at this exact head (`e672b4c`) in a clean clone, uncached: all 11 packages pass, including `internal/workflowcontract`.

### On the rewritten contract tests

Six tests asserted the matrix structure and could not survive the collapse. Rewriting tests so one's own change goes green is the failure mode to watch for here, so each rewritten assertion was checked by mutation — break the workflow, confirm the test fails:

1. removing the empty-tree guard -> `TestZshSyntaxFailsWhenNoFilesDiscovered` fails
2. stopping at the first failure -> `TestZshSyntaxReportsEveryFailure` fails on the tally
3. interpolating the filename into `zcompile` -> `TestZshCompileStepTreatsMetacharacterFilenameAsData` fails

Mutation 3 is the one worth noting: it genuinely executes the injected command (the `injected` marker file appears on disk) and the test observes that, so it tests the security property rather than the wording of the step.

Net, the suite is stronger than before. The newline-filename and metacharacter-filename properties are now proven by running the script over a real fixture tree instead of pattern-matching YAML, and the empty-tree guard is a property the matrix never had — with 58 separate jobs a `find` mistake was visible, whereas one collapsed job could silently check nothing and report success.

The `zsh-n-complete` aggregate job was removed rather than kept: it existed only to publish a stable context name in front of matrix legs whose names moved whenever a `.zsh` file was added or renamed. One job publishes `Zsh Syntax Check Complete` directly, and the test now pins that name, keeps the `pull_request` trigger unfiltered, and asserts no matrix strategy returns.

### Ruleset check

The branch ruleset requires exactly one context from this workflow, `Zsh Syntax Check Complete`, never the per-leg `zsh-n (<path>)` names. Verified by reading the ruleset. So no ruleset edit is needed and no required check is orphaned.

### Note on the stale red check

`Policy / Validate Branch Name` shows one FAILURE in the rollup from 04:56 and one SUCCESS from 05:14, both against commit `e672b4c`. The failure belongs to the superseded branch `ci-zsh-n-collapse` (now deleted), which pointed at the same commit; the success is this branch. GitHub scores the latest run per context, and `mergeStateStatus` is `CLEAN`. The stale row is cosmetic.

### Unrelated defect found while reviewing

Running the full suite repeatedly surfaced a pre-existing 15% flake in `TestAdapterCompositionAllFeaturesTogether`, which turned out to be a real parser false positive rather than noise. Filed as #337 with a three-line reproduction. It exists on `main` and is not caused by this PR.

### Assessment

Ready to merge. Scope is two files, the behavior is verified end-to-end rather than asserted structurally, and the one red check is a stale duplicate.

PR #338, opened 2026-09-21
fix(parse): disarm alternate-form scan on a non-brace while body (#337)

### LINKED ISSUE #337: Parser gap: adapter composition rejects valid Zsh when alternate if, while short form and try/always appear in that order
`TestAdapterCompositionAllFeaturesTogether` fails on roughly 1 in 7 runs on `main`, and the underlying cause is a real parser false positive, not just test noise.

Measured on `main` (`fcba451`), 60 uncached runs of that test alone:

```
flaked 9 / 60   (15%)
```

The test builds one file from `adapterSnippets`, which is a map, so Go randomizes the order on every run. The comment calls that "deliberate extra coverage ... since no ordering may matter" — but an ordering does matter, so the randomization turns a genuine defect into an intermittent failure.

## The defect

Delta-debugging a failing permutation shrinks it to three snippets:

```zsh
#!/usr/bin/env zsh
if (( 1 )) { x=1 }
while (( i < 3 )) (( i++ ))
{ true } always { true }
```

zsh accepts this:

```
$ zsh -f -n min3.zsh ; echo $?
0
```

zsh-lint rejects it:

```
$ zsh-lint min3.zsh
min3.zsh:4:1: `;` can only immediately follow a statement
exit=1
```

All three lines are individually valid and all 552 ordered *pairs* of snippets parse. The failure needs all three together, in that order: the alternate brace-form `if`, then an undelimited `while` short form, then a `try`/`always` block. Dropping any one of them makes it parse, so this is a composition defect in the adapter chain rather than a defect in any single adapter.

Randomized search over full-length permutations gives 13 failures in 4000 (0.3%), consistent with the observed per-run flake rate once you account for how many orderings put those three in the wrong relative position.

## Scope

Pre-existing on `main`, reproduced with a binary built from `origin/main` at `fcba451`. Not introduced by #326, #328 or #329, and unrelated to the CI change in #336 — that PR touches only `.github/workflows/zsh-n.yml` and `internal/workflowcontract/`.

## Suggested fix

Two separate things:

1. Fix the composition defect so the three-line file above parses.
2. Make the test deterministic. Iterating a map means a failure is unreproducible from the log alone: the printed source is the only record of the order, and a rerun usually passes, which invites treating a real defect as flake. Sorting the snippet names and asserting every permutation is too expensive at n=24, but sorting for the main assertion and adding a separate seeded randomized test that prints its seed on failure gives reproducibility and keeps the extra coverage.

ss-o:
Closes #337.

`TestAdapterCompositionAllFeaturesTogether` failed on roughly one run in seven. The test built its input by iterating a map, so Go's randomized order decided whether it passed, and the printed source was the only record of the order that failed. That read as flakiness; it was a real parser defect with a narrow trigger.

## The defect

`scanAlternateIfEdits` arms on the `while` keyword to recognize the alternate `while (( ... )) { ... }` form. Unlike `if`, a `while` may put its brace on the next line, so the scan searches past newlines for it. When the condition was followed by an ordinary body rather than a brace, nothing disarmed the state, so the scan kept going and claimed the next statement's brace as the loop body.

Three constructs had to meet:

```zsh
if (( 1 )) { x=1 }
while (( i < 3 )) (( i++ ))
{ true } always { true }
```

The `while` arm consumed the `{` of the `try`/`always` block, masking it into `; do true done always { true }`, and the retry was rejected at the `always` with an error naming a `;` the source never contained:

```
min3.zsh:4:1: `;` can only immediately follow a statement
```

`zsh -f -n` accepts the file. Both oracles agree it is a gap, per the parser-gap workflow. Each line parses alone and all 552 ordered pairs parse; only the three together reproduced it.

## The fix

An arithmetic condition not followed by `{` now clears the state before moving on. The alternate form is untouched, including a brace on the next line, which is reached through the arm that scans for it rather than through this fall-through.

## Verification

Differential run of a binary built before and after, over all 52 corpus fixtures, the 5 invalid `.txt` fixtures, and 7 targeted probes: exactly one input changes behavior, the new fixture, and it changes by accepting valid Zsh. Nothing else moves.

Flake rate on the previously flaky test, uncached:

| | before | after |
|---|---|---|
| `TestAdapterCompositionAllFeaturesTogether` | 9 / 60 (15%) | **0 / 120** |
| sampled full-length orderings | 13 / 4000 | **0 / 300** |

Each of the three new guards was mutation-tested by reverting the one-line fix:

- `TestParseAlternateIfDisarmsWhileWithoutBraceBody` fails
- `TestAdapterCompositionRandomOrders` fails, reporting `replay with seed 1`
- the corpus gate fails on `ok-alternate-if-while-try-always.zsh`

`go test -race -count=3 ./internal/parse/` passes, `gofmt` and `go vet` are clean, and the full suite is green.

## On the test change

The randomized coverage is kept, but made reproducible. The all-features test now iterates in sorted order, so a failure always reproduces. A companion test samples orderings from explicit seeds and prints the seed and order on failure.

The sample count is 40 rather than 200 deliberately: each parse of a 24-construct file walks the adapter chain, and 200 seeds pushed `internal/parse` from 0.3s to 15.9s. Forty seeds cost about 3s and still catch this defect on seed 1.

The old comment claimed map randomization was "deliberate extra coverage ... since no ordering may matter." An ordering did matter, and randomizing without recording the seed is what let a real defect sit as an accepted flake.


ss-o:
## Review

Self-review under the ADR-0026 fallback (Copilot rate-limited). GitHub refuses an approving review on one's own pull request, so this is the review of record as a comment.

### Scope

Four files: a one-line behavior change with a comment in `alternate_if.go`, a focused test, the composition-test determinism change, and one corpus fixture. The fix itself is `currentIf = ifNone` on a fall-through path.

### What I checked beyond "tests pass"

**Both oracles, per the parser-gap workflow.** `zsh -f -n` accepts the repro and `zsh-lint-survey` rejected it before the fix. A file both reject would be a broken script, not a gap; this is a gap.

**The fix is not over-broad.** The risk with disarming is silently narrowing the feature the adapter exists for. Differential run of before/after binaries across all 52 corpus fixtures, the 5 invalid `.txt` fixtures, and 7 targeted probes: exactly one input changes, the new fixture, and it changes by accepting valid Zsh. The alternate form still works, including `while (( i < 3 )) { ... }` and `while (( 1 )) && (( 2 )) { ... }`, asserted on tree shape rather than exit code.

**Each guard actually guards.** Reverting the one-line fix must break every test added; verified individually:

- `TestParseAlternateIfDisarmsWhileWithoutBraceBody` fails
- `TestAdapterCompositionRandomOrders` fails with `replay with seed 1`
- the corpus gate fails on `ok-alternate-if-while-try-always.zsh`

**Flake actually gone,** not merely unobserved: 0/120 uncached runs, against 9/60 measured before. Sampled orderings 0/300, against 13/4000.

### Two things worth a second opinion

**The fixture is deliberately combination-heavy.** Its value is in the pairings, not the lines, since every line parses alone. That makes it less readable than most corpus files; the header comment says so explicitly.

**Sample count is a tradeoff, not a rounded number.** 200 seeds pushed `internal/parse` from 0.3s to 15.9s, which is a real cost on every run. 40 seeds cost about 3s and still catch this defect on seed 1. If a future ordering defect is rarer than ~1 in 40, this test will miss it on a given run, though repeated CI runs across commits still sample the space.

### Note on the red check

`Policy / Validate PR Title` shows a FAILURE from 05:47 and a SUCCESS from 05:50. The first ran against the original title, which was 82 characters and exceeded the policy's 72-character limit on the description; it was shortened. GitHub scores the latest run per context, and `mergeStateStatus` is `CLEAN`.

### Assessment

Ready to merge. The root cause is understood rather than pattern-matched, the blast radius is measured rather than assumed, and the test change removes an accepted flake that was hiding a real defect.

PR #339, opened 2026-09-21
fix(parse): reject carriage returns Zsh rejects (#333)

### LINKED ISSUE #333: Parser gap: CRLF line endings are accepted although native Zsh rejects them
`zsh-lint` treats a trailing carriage return as whitespace and accepts CRLF files that native Zsh rejects. This is a **false accept**: the linter passes a script that will not run.

## Reproduction

A file with CRLF line endings and no loop at all is enough — this is not specific to any one construct:

```zsh
f() { print x }
```
written with `\r\n` line endings.

```
$ zsh -f -n crlf.zsh ; echo $?
crlf.zsh:1:15: parse error near `\r'
1

$ zsh-lint crlf.zsh ; echo $?
0
```

Native Zsh makes `\r` part of the preceding token, so `}\r` is not a closing brace. The parser (mvdan/sh) skips it as whitespace.

The mirror case also exists — a `\r` *inside* a header is a false **reject**:

```zsh
f() { while true\r }
```
is accepted by `zsh -f -n` (exit 0) and rejected by `zsh-lint`.

## Pre-existing, and how that was established

This is not a regression from #329. It was confirmed by building the survey binary at `ba9f3ca` (the base of #329) and comparing:

| source | `zsh -f -n` | base `ba9f3ca` | #329 head |
|---|---|---|---|
| `f() { print x }` + CRLF | 1 | **0** | 0 |
| `while true; do :; done` + CRLF | 1 | **0** | 0 |
| `f() { while true }` + CRLF | 1 | 1 | **0** |

The first two rows contain no empty loop body and already disagree at base, which is what shows the cause is CRLF handling rather than any loop construct. The third row previously failed for an unrelated reason — the empty body of #327 — so the CRLF defect was **masked** behind that error; fixing the empty body in #329 simply stopped hiding it.

## Suggested scope

Decide the intended contract first, since it is a policy question rather than a parser bug in isolation: either reject CRLF input outright with a clear diagnostic, or normalise line endings on read and keep positions mapped back to original offsets. The second is friendlier but every adapter's source map has to stay honest about the original byte offsets.

## Verification notes

Check reproductions as **files** with explicit `\r\n` bytes, not with `zsh -c`: `-c` supplies no trailing newline, which changes the verdict (the trap recorded in #287).

ss-o:
Closes #333.

Zsh lexes `\r` as an ordinary word character, not as whitespace. mvdan/sh skips it, so the linter accepted CRLF files that Zsh refuses to run — a **false accept**, reporting success for a script that cannot run. A file with no loop at all was enough:

```
$ zsh -f -n crlf.zsh ; echo $?
crlf.zsh:1:15: parse error near `\r'
1

$ zsh-lint crlf.zsh ; echo $?
0
```

## The rule, established by probing rather than by reading the grammar

The fix is **not** "reject any file containing `\r`". Zsh accepts a `\r` wherever a word may appear, and rejecting those would trade a false accept for a false reject. Verified as files against `zsh -f -n`:

| source | zsh | why |
|---|---|---|
| `x=1\r` | valid | `\r` may be a word |
| `print x\r` | valid | argument position |
| `print x;\r` | valid | after `;`, command position |
| `true &&\r` | valid | after an operator |
| `print 'a\r'` | valid | quoted |
| `f() { print x }\r` | **error** | `}` takes no argument |
| `while true; do :; done\r` | **error** | `done` takes no argument |
| `if true; then :; fi\r` | **error** | `fi` takes no argument |
| `case x in y) :;; esac\r` | **error** | `esac` takes no argument |
| `(true)\r` | **error** | `)` takes no argument |

A space before the `\r` does not help, which is what rules out "the `\r` glues onto the preceding token" as the explanation. The actual cause is simpler: `} extra` and `done extra` are already errors here, and `} \r` is the same shape.

## Approach

The verdict comes from re-parsing with `\r` replaced by an ordinary word byte, not from a hand-written position rule, so the guard cannot drift from the front end as adapters are added. The substitution is one byte wide, so every offset is preserved.

The probe's own error is **discarded**. It names whatever construct the extra word broke — typically the enclosing `while` or `if` — which would send a reader to the wrong line:

```
empty-body.zsh:1:7: `while <cond>` must be followed by `do`      <- misleading
empty-body.zsh:1:19: parse error near `\r`                       <- reported
```

The diagnostic is reported at the `\r` itself, reusing Zsh's own wording, following the precedent of `rejectCloseBraceWords`.

## Verification

- **45 probe sources** covering both sides of the rule: **45/45 agree** with `zsh -f -n`, against 8 disagreements before.
- **Differential before/after** over 52 corpus fixtures, the existing invalid fixtures, and **40 real `.zsh` files** in the workspace: no other behavior change.
- **Mutation test**: removing the guard from `Parse` fails 12 subtests.
- `go test -race -count=2 ./internal/parse/` clean; `gofmt` and `go vet` clean; full suite green.
- **Cost**: the probe runs only when the source contains a `\r`, so a 164 KB CR-free file costs 2 ms more (357 ms -> 359 ms).

## Note on the mirror case

The issue also mentions `f() { while true\r }` as a false *reject*. That one is now handled by the same rule — the `\r` sits where a word is allowed, the probe parses, and the guard stays silent. It is covered in `TestParseAcceptsCarriageReturnZshAccepts`.


ss-o:
## Review

Self-review under the ADR-0026 fallback (Copilot rate-limited). GitHub refuses an approving review on one's own pull request, so this is the review of record as a comment.

### Why this one matters more than its size suggests

A false accept is the worst defect class for a linter: it reports success for a script that cannot run. Everything else this repository has fixed recently was a false *reject*, which is annoying but visible. This one was silent.

### What I checked beyond "tests pass"

**The rule was derived by probing, not by reading the grammar.** My first hypothesis — that `\r` glues onto the preceding token, so `done\r` is no longer the reserved word — was **wrong**, and `done \r` (with a space) disproved it by failing too. The actual rule is that `\r` is an ordinary word character, and a word is an error after `}`, `done`, `fi`, `esac`, `)`. Confirmed by `} extra` failing identically to `} \r`.

That mattered for the fix: a blanket "reject `\r`" rule would have broken `x=1`, `print x`, `true &&`, and quoted `\r`, all of which are valid CRLF Zsh. 45 probe sources now agree with `zsh -f -n`, against 8 disagreements before.

**The blast radius is measured, not assumed.** Before/after binaries diffed over 52 corpus fixtures, the existing invalid fixtures, and 40 real `.zsh` files from the workspace: zero unintended changes.

**The guard actually guards.** Removing the call from `Parse` fails 12 subtests.

**Cost is bounded.** The probe only runs when the source contains a `\r`, so the common path pays one `bytes.IndexByte`. A 164 KB CR-free file: 357 ms -> 359 ms.

### Design choice worth a second opinion

Deciding the verdict by **re-parsing with a substituted byte** rather than by a hand-written position rule is the main judgment call. The argument for it: the set of positions where a word is legal is exactly what the front end already encodes, and it grows every time an adapter lands. A hand-written rule would need updating in lockstep and would silently drift. The cost is one extra parse for CRLF files only.

The alternative — enumerate the closers and check the following byte — is cheaper but would have to re-derive `closeBraceExemptWord`-style exceptions (assignment values, case patterns, here-doc bodies) by hand. I judged drift the bigger risk. Worth a maintainer disagreeing.

### Assessment

Ready to merge. 14/14 checks green, `mergeStateStatus: CLEAN`, no stale rows this time.

Issue #330, opened 2026-09-21
Parser gap: a while/until condition list continuing on the next line is rejected


ss-o:
Native Zsh accepts a `while` or `until` header whose condition list continues on the **following line**, and `zsh-lint` rejects it.

```zsh
while true
print x
```

`zsh -f -n` accepts it (exit 0). `zsh-lint` reports:

```
t.zsh:1:1: `while <cond>` must be followed by `do`
```

## This is an empty body, not a next-line body

The original text of this issue claimed the next line "really becomes the body". That is **wrong**, and it was disproved by executing the forms rather than reading the grammar. Correction recorded in the comment thread; the body below is the corrected spec.

If `print X` were the loop's **body**, a false condition would run it zero times:

```
$ printf 'while (( 0 ))\nprint X\n' > a.zsh; timeout 2 zsh -f a.zsh
X
X
X ...        # infinite, so print is NOT the body
```

And a true condition with `false` on the next line **terminates**, so the next line's exit status is what ends the loop:

```
$ printf 'while (( 1 ))\nfalse\n' > b.zsh; timeout 2 zsh -f b.zsh; echo $?
0
```

Control, with a real body — a false condition is correctly silent:

```
$ printf 'while (( 0 )); do print X; done\n' > c.zsh; timeout 2 zsh -f c.zsh
             # no output
```

So the newline-separated sublist joins the **condition list**, and the body is **empty**. `while (( 0 ))` / `print X` means `while (( 0 )); print X` with nothing to iterate: the loop's status comes from the last element of the condition list, which is why `print` (always succeeding) loops forever and `false` stops it.

## Scope of the fix

This is the #327 empty-body family with a condition list that crosses a newline, **not** a separate next-line-body feature. The fix is to let the existing empty-body adapter accept a condition list spanning newlines, rather than to add a new adapter that places `done` after the next sublist.

Shapes to cover (each verified with `zsh -f -n` as a file):

```zsh
while true
print x
```

```zsh
until true
print x
```

```zsh
h() { while (( 1 )); print x }
```

```zsh
while true
print a
print b
```

## Related

- #327 / PR #329 — the empty-body adapter this should extend.
- #332 / PR #340 — its recorded limitation ("a bare loop followed by another statement is rejected") is this same shape: the neighbouring statement is pulled into the loop's condition list.

ss-o:
## Correction to this issue's premise — filed by me, and wrong

[...]

PR #340, opened 2026-09-21
fix(parse): accept a redundant `;` in command position (#332)

### LINKED ISSUE #332: Parser gap: a redundant semicolon before an empty loop body is rejected
Native Zsh accepts a redundant `;` between a loop condition and its empty body. `zsh-lint` rejects it.

## Reproduction

```zsh
while true; ;
```

```
t.zsh:1:13: `;` can only immediately follow a statement
```

`until true; ;` behaves the same way.

## Why it is filed separately

PR #329 (#327) normalises the separators between a condition and an empty body: it keeps one `;`, blanks any further ones, and inserts `; do` only when none is present. That logic handles a single separator plus newlines, which is what the parser accepts before `do`. A bare `;` standing alone as its own token is rejected by the parser at its own offset before any of that applies, so it needs separate handling rather than an extra case in the existing scan.

Low priority: the shape is legal but has no practical use. Filed so the gap is recorded rather than rediscovered.

## Verification notes

Check reproductions as **files**, not with `zsh -c`: `-c` supplies no trailing newline, which changes the verdict (the trap recorded in #287).

Related: #327 (empty body, fixed in #329).

ss-o:
Closes #332.

Zsh reads a list as zero or more sublists, so a `;` standing in command position terminates an empty sublist and does nothing. Confirmed by **running** it rather than by reading the grammar:

```
$ zsh -f -c 'print a; ;'          # prints a, once
$ zsh -f -c 'i=0; while (( i++ < 2 )); ; do print loop; done'   # loops twice
```

## The gap is wider than the issue reports

The issue describes `while true; ;`. Probing showed the same rejection for shapes with no loop at all:

| source | zsh | before |
|---|---|---|
| `;` | valid | rejected |
| `print x; ;` | valid | rejected |
| `; ; ;` | valid | rejected |
| `print run; ; print again` | valid | rejected |
| `while true; ;` | valid | rejected |
| `f() { print body; ; }` | valid | rejected |

All of them share one parser complaint, so one adapter covers the class rather than the reported instance.

## Design

The `do` (#238) and `then`/`else` (#297) adapters already handle this byte **after a reserved word**, where the keyword anchors the site. A redundant `;` in plain command position has no anchor, so this adapter blanks every site in one pass and lets the re-parse decide — an empty sublist owns no node, so a tree that parses is the tree the source describes.

**One pass, not one site per re-entry.** Masking a single occurrence and recursing costs a full re-parse per separator, which is quadratic in a file that uses the form repeatedly (the lesson from #327). Measured flat:

```
n=  2  120µs     n=  8   42µs     n= 32  101µs
```

and pinned by a test that counts the retry parser's calls (`want 1`).

## Two findings worth flagging

**A case terminator must be stepped over whole.** My first version read `;;` one byte at a time, so in `case x in y) ; ;; esac` the second `;` looked like an empty sublist and got masked — silently deleting the terminator. Caught by my own test, now mutation-pinned.

**I removed a guard I had written.** I originally checked that the parser's reported offset was one of the scanner's sites. Probing for an input that would exercise it found none: whenever the error text is ours, the reported `;` is a site by construction (a `;` the parser *can* place is an ordinary terminator; a `;;` it cannot is reported under different text). An unreachable guard no test can pin is worse than none, so it is gone and the reasoning is a comment.

## Verification

- Every source checked against `zsh -f -n` **as a file**; the executable ones were run to confirm the `;` is a no-op, not merely accepted.
- **Differential before/after** over the corpus, the invalid fixtures, and 40 real `.zsh` files: **exactly one verdict changes** — the new fixture.
- **Mutation-tested**: treating case terminators as sites fails `redundant_then_terminator`.
- New corpus fixture is a real guard: native-valid, rejected by the base binary, accepted after.
- `gofmt`, `go vet`, full suite green.

## Deliberately still rejected

Two shapes are **not** this bug, and both were confirmed to behave identically before this change:

- `true && ;` — fails under a different parser error.
- A bare loop followed by another statement (`while true; ;` then `print x`) — this is **#330**, where the next statement becomes the loop's body. It is why the fixture keeps a `}` or end-of-file close to each bare loop, and it means #332 cannot be fully closed for the loop-plus-statement case until #330 lands.


ss-o:
## Review

### Scope call

The issue reports `while true; ;`. I widened the fix to the class — a file opening with `;`, `print x; ;`, `; ; ;` — because probing showed they all produce the same parser complaint. Fixing only the reported instance would have left an adapter that declines its own error text on neighbouring inputs, which is how chains accumulate near-duplicate entries.

### Two things I changed my mind about mid-implementation

**A case terminator must be stepped over whole.** My first scanner read `;;` byte by byte, so in `case x in y) ; ;; esac` the leading `;` closed the body, `commandStart` went true, and the second `;` of the terminator was flagged as an empty sublist. Masking it would have silently deleted a case terminator — a meaning change, not a parse failure, so the corpus would not necessarily have caught it. My own unit test did.

**I deleted a guard I wrote.** I had checked that the parser's reported offset was one of the scanner's sites, reasoning it would stop the adapter masking an unrelated defect. Mutation testing said the check was untested; rather than write a test for it I went looking for *any* input that would exercise it, and there is none: when the error text is ours the reported `;` is always a site by construction. A `;` the parser can place is an ordinary terminator, and a `;;` it cannot place is reported under different text (`` `;;` can only be used in a case clause ``). I removed it and left the reasoning as a comment. An unreachable guard that no test can pin is worse than no guard — it reads as protection that is not there.

### Honest limits

`while true; ;` **followed by another statement** is still rejected. That is #330 (the next statement becomes the loop's body), confirmed identical on the base binary. It shapes the fixture: bare loops there keep a `}` or end-of-file close by. So #332 is closed for the general case but the loop-plus-statement combination stays broken until #330 lands. I would rather say that plainly than ship a fixture engineered to look complete.

`true && ;` is also still rejected, under a different error. Filed separately rather than widened into this PR.

### Verification I trust

- Sources verified as **files** against `zsh -f -n`, and the executable ones **run** — `print a; ;` prints `a` once. Accepting a parse is weaker evidence than observing the no-op.
- Differential before/after over corpus + invalid fixtures + 40 real `.zsh` files: **exactly one verdict changes**, the new fixture.
- The one-pass claim is measured (flat 2 -> 32 separators) *and* pinned by a call-count test, so a future refactor to per-site recursion fails the suite rather than quietly regressing.
- 14/14 CI green, `mergeStateStatus: CLEAN`.

### Assessment

Ready to merge.

PR #341, opened 2026-09-21
fix(parse): read a while or until condition as a list (#330)

### LINKED ISSUE #330: Parser gap: a while/until condition list continuing on the next line is rejected

ss-o:
Closes #330.

Native Zsh reads a `while` or `until` header's condition as a **list**, so it does not stop at the first statement — it absorbs every sublist up to the end of the enclosing list, and the loop's status is that list's last element. The adapter ended the condition at its first statement, so `while true` followed by anything on the next line was rejected.

## The issue's original premise was wrong, and I wrote it

This issue claimed the next line "really becomes the body". Executing the forms disproves it; the body has been corrected and this PR implements the corrected spec.

If the following sublist were the **body**, a false condition would run it zero times:

```
$ printf 'while (( 0 ))\nprint X\n' > a.zsh; timeout 2 zsh -f a.zsh
X
X ...        # infinite => the sublist is in the CONDITION
```

And a true condition with `false` next **terminates**, so the last element is what ends the loop:

```
$ printf 'while (( 1 ))\nfalse\n' > b.zsh; timeout 2 zsh -f b.zsh; echo $?
0
```

Control, with a real body — correctly silent:

```
$ printf 'while (( 0 )); do print X; done\n' > c.zsh; zsh -f c.zsh
             # no output
```

A `;` on one line behaves identically (`while (( 0 )); print X` loops forever). Only the no-separator short form (#211) and the brace form give a real body.

## Consequence for the fix

This is the **#327 empty-body family** with a condition list crossing a separator, not a separate next-line-body construct. So it extends the existing adapter rather than adding one: `statementListEnds` widens each candidate's condition to the end of the list that holds it. Probe, retry and verification are unchanged.

It also closes the limitation PR #340 (#332) recorded — "a bare loop followed by another statement is rejected" — because that statement was always part of the condition list.

## Verification

- **Sources checked as files** against `zsh -f -n`; the decisive ones **executed**, since parsing cannot distinguish a body from a condition.
- **Differential before/after** over the corpus, invalid fixtures and 60 real `.zsh` files: **exactly one verdict changes**, the new fixture.
- **Adversarial**: 8 sources placing a real syntax error after the header (`{`, `fi`, `done`, `if`, unclosed quote, `)`, `esac`, `do`) all **stay rejected** — widening the condition does not swallow errors.
- **The tree is asserted, not just the parse**: condition length, empty body, `Until`, and that two bare loops nest (outer cond=3 holding inner cond=2).
- **Mutation**: disabling the widening fails all 5 shape subtests plus 6 parse cases.
- `-race` clean, `gofmt`/`vet` clean, full suite green.

## A constraint worth naming

The corpus fixture **terminates under `zsh -f`**, verified by running it. That shapes its layout: a bare loop's condition swallows every later sublist in its list, so two in one list can never both halt — the first re-runs the second forever. Each bare loop in the fixture is bounded by a closer (function body, subshell, `do ... done`) and the single top-level one comes last.

I found this by running the fixture rather than assuming, after writing a comment claiming "nothing here loops" that was false. The unbounded shapes are asserted in `while_condition_list_test.go`, where only parse and tree shape matter and nothing is executed.


ss-o:
## Review

### The part that deserves scrutiny

This widens the condition of every `while`/`until` candidate to the end of its enclosing list. That is a broad change to a core path, and the obvious risk is swallowing a genuine syntax error that happens to sit after the header — turning a false reject into a **false accept**, which is strictly worse.

I tested that directly rather than reasoning about it: 8 sources placing a real error after the header (`{`, `fi`, `done`, `if`, unclosed quote, `)`, `esac`, `do`) all stay rejected. The reason they hold is that `selectEmptyBodyAt` still has to agree the body byte ends a list, and `verifyWhileEmptyBodies` still requires a `WhileClause` with an empty `Do` at every site — the widening only moves where the condition ends, not what counts as a valid result.

### Why I assert the tree, not just the parse

Parsing cannot distinguish "condition list" from "body" — both produce a tree. Only execution can, which is why `TestParseWhileConditionListShape` pins condition length and an empty body, and why the nested case asserts outer cond=3 holding inner cond=2. A test that only checked `err == nil` would have passed just as happily on a wrong tree.

### Corrections I made to my own work

**The issue's premise was mine and it was wrong.** I filed #330 claiming the next line becomes the body, with reasoning about why it needed handling separate from #327. Executing it showed the opposite: the sublist joins the condition, the body is empty, and this is #327's family. The fix is consequently ~15 lines extending an existing adapter rather than the new adapter I had specified.

**I wrote a fixture comment that was false.** It claimed "nothing here loops at run time"; running it showed it hung. Two further attempts to fix the layout also hung, each for a different reason, before I isolated the actual rule — a bare loop's condition swallows every later sublist, so two in one list can never both halt. The fixture is now verified terminating by execution, and the unbounded shapes live in the unit test where nothing runs.

**A mutation test reported "0 failures" and I nearly believed it.** The first attempt deleted a variable's only use, so the package failed to *compile* — a build failure proves nothing about tests. The second used a `grep -c` filter that printed nothing. Only the unfiltered run showed the mutation genuinely fails 5 shape subtests and 6 parse cases. Second time this session I have been misled by my own output filtering; both instances are now in the skill.

### Assessment

Ready to merge. One verdict changes across corpus + fixtures + 60 real files, and it is the intended one.

Issue #331, opened 2026-09-21
Parser gap: a loop condition ending in a trailing && or || is rejected


ss-o:
Native Zsh accepts a `while` or `until` header whose condition ends in a trailing `&&` or `||` operator with nothing after it. `zsh-lint` rejects it.

## Reproduction

```zsh
x=0
while (( x )) &&
```

`zsh -f -n` accepts it (exit 0). `zsh-lint` reports:

```
t.zsh:2:1: `while <cond>` must be followed by `do`
```

A trailing `|` behaves the same way.

## Why it is filed separately

PR #329 (#327) handles the empty body that native Zsh reads *before* a `|`, `||` or `&&` operator that takes the loop as its **left** operand — `w() { while true | cat }` and `v() { while true || print x }` are both accepted there. This issue is the mirror image: the operator is the **last** token of the header, so there is no right operand, and the empty sublist sits after a dangling operator rather than before a complete one. The adapter's probe reads the condition through `&&`/`||` chains to find where it ends, which is why a dangling operator needs its own handling.

## Verification notes

Check reproductions as **files**, not with `zsh -c`: `-c` supplies no trailing newline, which changes the verdict (the trap recorded in #287).

Related: #327 (empty body, fixed in #329).

ss-o:
### Template fields missed at filing

This was filed without the `Parser gap` issue form, so two required fields were omitted. Recording them here rather than rewriting the body, which now carries discussion.

- **Zsh manual section:** zshmisc, Lists — `&&` and `||` join sublists; a trailing operator continues the list onto the next line
- **zsh-lint version:** v1.3.0 (latest release; reproduced on `main` at the time of filing)
- **Where it occurs:** Not yet investigated in depth; filed from adversarial probing.

Title and labels (`parser-gap`, `type:bug`) have been corrected to match `.github/ISSUE_TEMPLATE/parser-gap.yml`.

ss-o:
### Still open after #341, and one correction

Re-checked against `main` at PR #341 (#330), which widened a `while`/`until` condition to the end of its enclosing list. That did **not** fix this issue — a *dangling* operator is a different shape from a condition list that continues.

```
x=0
while (( x )) &&          # zsh -f -n: exit 0; zsh-lint: `&&` must be followed by a statement
```

Confirmed still a gap for `&&` and `||`, and inside a function body (`h() { while true &&` / `}` → `` `}` can only be used to close a block ``).

### Correction to this issue's text

[...]

Issue #324, opened 2026-09-20
Parser gap: multi-name for with a parenthesized word list (for a b (1 2) { : }) is rejected


ss-o:
## Summary

The multi-name for loop with a parenthesized word list, `for name ... ( word ... ) sublist`, is rejected in every body form: the for adapter (#13) rewrites `for a (1 2) { : }` and `for a b in 1 2; do :; done`, but `for a b (1 2) { : }` keeps the parser's header error at the `for` word. Native Zsh accepts every row and `functions f` shows it reads the paren list as the `in` list of a two-name loop.

## Minimized reproduction

Every row has empty `zsh -f -n` stderr (zsh 5.9.2).

| Source | zsh-lint `main` after #322 |
| --- | --- |
| `for a b (1 2) { : }` | `` 1:1: `for foo` must be followed by `in`, `do`, `;`, or a newline `` |
| `for a b (1 2 3 4) { print $a $b }` | same |
| `for a b (1 2); { : }` | same |
| `for a b (1 2) do :; done` | same |
| `for a b (1 2) :` | same |
| `for a b c (1 2 3) { : }` | same |
| `for a (1 2) { : }` | parses |
| `for a b in 1 2; do :; done` | parses |
| `for a b in 1 2; { : }` | `` 1:17: for loops with braces are a bash/mksh feature; tried parsing as zsh `` (#261) |

`functions f` on `f() { for a b (1 2 3 4) { print $a $b } }` prints `for a b in 1 2 3 4` with a `do ... done` body, so the paren list is the plain word list of the multi-name loop.

## Classification

[Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Complex-Commands): `for name ... [ in word ... ] term do list done`, with more than one name; [Alternate Forms For Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Alternate-Forms-For-Complex-Commands): `for name ... ( word ... ) sublist`. The two productions compose, and the front end handles each on its own but not together.

## Corpus evidence

None in the surveyed z-shell corpus; found while probing #321.

## Acceptance criteria

- A `gap-<id>-multi-name-paren-for.zsh` fixture with the failing rows fails on the public parse path before the fix.
- Each row parses into the same `ForClause` the `for a b in 1 2; do ... done` form gives, with the names, items and body statements at their original positions.
[...]

PR #342, opened 2026-09-21
docs(project): record the 2026-09-21 survey and pin two fixtures

ss-o:
Records the survey run that four merged parser PRs owed, and pins two unprotected fixtures.

## Why

`docs/project/parser-gap-workflow.md` step 5 requires re-running the discovery survey after any adapter change:

> Because the survey reports only the first error per file, this class of defect masks real gaps rather than merely reporting false ones. **Re-run the discovery survey after any adapter change and expect the reported set to shift.**

#338 (#337), #339 (#333), #340 (#332) and #341 (#330) all landed without one. Each was verified against unit tests and before/after binary differentials, but never against the reference corpus. This closes that gap.

## Results at the pinned revisions

18-file set resolved from `corpus-paths.txt` across the six repositories at `corpus-revisions.txt`:

| Check | Result |
|---|---|
| native `zsh -f -n` | 18/18, 0 failures |
| parser survey | **18 ok, 0 failed** |
| analyzer `--no-config` | 0 errors, 0 warnings (exit 0) |

**Nothing was unmasked** — the reported set did not shift. One advisory Info finding (`plugin/unload-function` in `zsh-fancy-completions/lib/widgets.zsh:24:3`) is advisory by contract and does not fail the gate.

## Drift check

Ran the consumer-`main` pass too, since that is the drift signal. Three consumers have moved:

| Repository | Pinned | Consumer `main` |
|---|---|---|
| `src` | `d19ae48` | `d262ede` |
| `zd` | `35b837d` | `5d16059` |
| `z-a-meta-plugins` | `ba34790` | `74bd8a8` |

Still 18 files, still native-valid, still **18 ok / 0 failed**. No revision bump is due on parser evidence.

## The finding worth acting on

**The corpus is exhausted as a source of parser gaps at these revisions.** Every gap fixed this session (#330, #332, #333, #337) came from adversarially probing the Zsh grammar, not from corpus evidence — and so did the open ones (#331, #324, #307).

That leaves step 2's "prioritize by corpus evidence" with no input. Either the corpus expands (#138 / #132) or priority comes from manual probing. The record states this explicitly so a later reader does not take `0 failed` for `no work to do`.

## Fixture baseline

`requiredFixtures` guards against accidental deletion. `ok-redundant-separator.zsh` (#332) and `ok-while-condition-list.zsh` (#330) were added this session and left out of it — step 5 says to update the list and I didn't.

Verified non-vacuous rather than assumed: removing either fixture now fails `TestMinimizedCorpus` with `required baseline fixture missing`.

## Verification

- Corpus checked out at the exact pinned revisions, restored to those pins afterwards.
- Directory entries expanded with NUL-delimited `find -type f`, matching the workflow, so dot-prefixed extensionless function files are included.
- `gofmt`, `go vet`, full suite green.
- Docs-only plus one test-data list; no behavior change.


ss-o:
## Review

### Scope

Docs-only plus one test-data list. No parser or analyzer behavior changes; the `internal/survey/corpus_test.go` edit adds two entries to `requiredFixtures`.

### What was verified

- Corpus checked out at the exact revisions in `corpus-revisions.txt`, and restored to those pins after the drift pass.
- File set resolved the same way the gate does (NUL-delimited `find -type f`), giving 18 files including dot-prefixed extensionless function files.
- Native `zsh -f -n` accepts all 18 at both the pins and consumer `main`.
- Parser survey: 18 ok / 0 failed at both.
- Analyzer `--no-config`: exit 0, one advisory Info finding that does not fail the gate.
- `requiredFixtures` guard proven non-vacuous by removing each new fixture and observing `required baseline fixture missing`.
- `gofmt`, `go vet`, full suite green.

### Judgment

The valuable output is not "everything passes" — it is that **the corpus can no longer produce a parser gap at these revisions**, which invalidates step 2's prioritization input. The record states that explicitly rather than leaving a future reader to infer that `0 failed` means there is no work left.

Prose wrapping matches the existing `docs/project/` convention (~70-75 char median, nothing over 90); Prettier runs without config, so `proseWrap: preserve` leaves it alone and the convention is hand-maintained.

Merging.

PR #343, opened 2026-09-21
docs: use semantic line breaks in prose

### LINKED ISSUE #344: Use semantic line breaks in Markdown prose
## Problem

Markdown prose in this repository is hard-wrapped to a column.
That is not neutral on GitHub.

GitHub renders Markdown with GFM line breaks, so **every newline inside a paragraph becomes a `<br>`**.
Column-wrapped prose therefore renders with breaks at arbitrary mid-sentence positions.

Verified against GitHub's `/markdown` endpoint, using text committed in `docs/project/parser-gap-workflow.md`:

```html
<p>Because the survey reports only the first error per file, this class of defect<br>
masks real gaps rather than merely reporting false ones. Fixing it immediately<br>
exposed a further genuine gap in the same file.</p>
```

Three `<br>` tags in a four-line block, every one mid-sentence.

## Proposal

Use semantic line breaks: one sentence per line, breaking a long sentence at a clause boundary.

The breaks land at sentence ends, where they read as intended.
Editing a sentence also touches only its own line, so diffs stay reviewable instead of reflowing a whole paragraph.

## Scope

The living contracts, the two `.github/instructions/` files, and the current survey.

Dated survey reports from earlier days stay untouched: `docs/project/README.md` states they are history and older ones are not updated.

Tables, code fences, link definitions, and frontmatter are out of scope.

## Notes

Nothing enforces this today, and this proposal does not add enforcement.
`prettier` runs unconfigured, so `proseWrap: preserve` leaves prose alone, and `markdownlint` runs with MD013 off — a 989-character line currently passes both.
The convention belongs in `AGENTS.md` and in review.

Verification for any conversion should be that the rendered HTML is unchanged once `<br>` placement is normalized, proving only break position moved.

ss-o:
Closes #344.

Markdown prose here was hard-wrapped to a column.
That renders badly on GitHub, and this switches it to semantic line breaks: one sentence per line.

## The evidence

GitHub renders Markdown with GFM line breaks, so **every newline inside a paragraph becomes a `<br>`**.
I verified this against GitHub's own `/markdown` endpoint rather than assuming it.

The existing wrapped prose in `parser-gap-workflow.md` renders like this today:

```html
<p>Because the survey reports only the first error per file, this class of defect<br>
masks real gaps rather than merely reporting false ones. Fixing it immediately<br>
exposed a further genuine gap in the same file.</p>
```

Three `<br>` tags, every one mid-sentence.
The semantic-break version puts them at sentence ends instead, where they read as intended.

So this is not a cosmetic preference: the current style actively produces worse rendered output, and semantic breaks also keep diffs tight because editing a sentence touches only its own line.

## Scope

13 files: the living contracts, the two `.github/instructions/` files, and today's survey.

Dated survey reports from earlier days are **excluded** — `docs/project/README.md` states they are history and older ones are not updated.

## Verification

Each file was checked by rendering the before and after through GitHub's GFM endpoint, stripping the `<br>` tags that encode break position, and comparing the remainder. **All 13 render identically**, proving only break placement moved and no paragraph, table, code block, or inline markup changed.

That check earned its keep: it caught a nested list being flattened into its parent item (`- Nested follows: - Inner item`) when I first extended the converter to handle list bodies. The verifier was itself tested in both directions first — it reports SAME for a pure re-wrap and DIFFERENT when a single word changes.

Also confirmed:

- `trunk check --filter=prettier,markdownlint` clean on all 13. Prettier does not reflow prose (`proseWrap: preserve` by default, no config), and MD013 is off — I checked that a committed 989-character line currently passes.
- Go suite unaffected: 11 packages ok.
- Content is unchanged throughout; only line breaks move.

## Convention

Recorded in `AGENTS.md` under Documentation, including the fact that **nothing enforces it** — it is maintained by hand and in review. Tables, code fences, link definitions, and frontmatter are explicitly out of scope.


ss-o:
## Review

### Scope

Docs-only: 13 files, line breaks moved, no content changed. No code touched; the Go suite is unaffected (11 packages ok).

### The claim this rests on

"Renders identically" is the load-bearing claim, so it was measured rather than assumed. Each file's before and after were rendered through GitHub's own `/markdown` GFM endpoint, the `<br>` tags that encode break position stripped, whitespace collapsed, and the remainder compared. All 13 match.

The verifier was validated in both directions before use: SAME for a pure re-wrap, DIFFERENT when one word changes. A check that cannot fail proves nothing.

It then caught a genuine defect — extending the converter to list bodies flattened a nested list into its parent item. Fixed by treating any nested list marker as a new item rather than continuation prose, and re-verified.

### Judgment

The motivation is stronger than style preference. The previous convention produced mid-sentence `<br>` tags in rendered output on the primary reading surface, which is a real defect, not a matter of taste. Semantic breaks fix that and make future prose diffs line-scoped.

The empty retrigger commit is an artifact of `Validate Issue Link` reading the PR body from the event payload: editing the body does not re-fire the event, and `gh run rerun` replays the stale payload. Squash-merging collapses it.

Merging.

PR #345, opened 2026-09-21
fix(parse): accept a list ending in a dangling && or ||

### LINKED ISSUE #331: Parser gap: a loop condition ending in a trailing && or || is rejected

ss-o:
Closes #331.

Native Zsh accepts `&&` and `||` as the last token of a list.
The operator has no right operand, and the left one runs exactly as a bare statement would — `print a &&` prints `a`.

## The gap is wider than filed

I filed this as a loop-condition bug. It isn't: a dangling operator is valid wherever a list may end, so these were all rejected too.

```zsh
h() { print a &&
}
( print a &&
)
if true; then print a &&
fi
while true; do print a && done
case x in y) print a && ;; esac
```

## The rule, established by execution

An operator is only dangling at a **terminator**. Across a bare newline it takes the following statement as its right operand and short-circuits it:

```zsh
print first ||
print second
print third
```

prints `first` and `third` — `second` is the `||`'s right operand, skipped.

So the adapter confirms the operator by what follows it rather than trusting the error, and leaves every operator that has a real right operand untouched. That distinction is load-bearing: masking one would silently change what a script does, not merely whether it parses.

This also corrected the fixture twice. My first version printed `command not found: h`, because a preceding dangling `||` had swallowed the function definition — the very behaviour above, caught by running the fixture rather than only parsing it.

## `|` is excluded, deliberately

The issue body claimed a trailing `|` behaves the same way. It does not — Zsh rejects a dangling pipe, and a fix treating them alike would have traded a false reject for a **false accept**. Correction posted on #331 before implementing.

## Implementation

Two parser reports describe the same source shape: the operator itself when it ends the outermost list, or the closer when one follows, with wording that varies by keyword (`can only be used to close a block`, `can only be used in an \`if\``). Both are handled.

The operator is replaced with `; `, not blanks. Masking `do print a && done` with spaces leaves `print a done`, where `done` becomes an argument — caught by a mutation test.

Registered last in `adapterChain`, so keyword-anchored adapters keep precedence. A dangling operator in a `while` header leaves an empty loop body, which the #327 adapter then handles via `parseWithAdapters`.

## Verification

- **258 sources checked against `zsh -f -n`, zero disagreements**, including a generated grid of 216 prefix/operator/terminator combinations and 10 adversarial broken scripts that must stay rejected.
- Differential over corpus + invalid fixtures + 60 real-world files: **exactly one verdict changes**, the new fixture.
- Mutation tested. Removing the closer path fails 4 subtests; changing the `;` mask back to spaces fails `closer_same_line_done`; disabling the operator lookahead fails `operator_then_pipe`.
- Two mutations survived. I chased both rather than leaving them: an `if`/`elif` guard I had added turned out to change no verdict on 258 sources, so I **deleted it** — an unreachable guard reads as real protection. The remaining lookaheads are documented as defense in depth with the measurement stated, since an operator holding a right operand parses cleanly and never reaches an adapter.
- Fixture terminates under `zsh -f` (all 7 statements run), is rejected by base, accepted after, and is registered in `requiredFixtures` — proven non-vacuous by removing it.
- `gofmt`, `go vet`, full suite, and `-race -count=2` all clean.


ss-o:
## Review

### Risk

This adapter **edits source and re-parses**, so the failure mode that matters is not a missed fix — it is masking an operator that has a real right operand, which changes what a script *means* rather than whether it parses.

That is why the scanner confirms every candidate by what follows it, and why `TestDanglingAndOrLeavesCompleteOperators` asserts the operator survives in the printed tree rather than merely that the file parses.

### Scope correction

Filed as a loop-condition bug; it is not. A dangling operator is valid wherever a list may end, so function bodies, subshells, `then` branches, loop bodies and case arms were all affected. Fixing only the reported shape would have left most of the gap open.

### What the evidence covers

- 258 sources vs `zsh -f -n`, zero disagreements, including a generated 216-case grid and 10 deliberately broken scripts that must stay rejected.
- Differential: exactly one verdict changes.
- Mutation testing found two real defects — the `; ` mask (spaces leave `print a done`) and the closer path — and surfaced two survivors I chased rather than ignored.

### Judgment calls

I **deleted** an `if`/`elif` guard I had written once measurement showed it changed no verdict on 258 sources. The remaining lookaheads are documented as defense in depth with that measurement stated, instead of being presented as the thing holding the line.

I did **not** extend this to `|` despite the issue saying it behaves the same. It does not, and doing so would have introduced a false accept.

The fixture needed two corrections, both caught by running it: a preceding dangling `||` was swallowing later statements — the same short-circuit behaviour the fix is about.

Rebased onto `fbb5d3b` and lint-clean after applying De Morgan's law to two conditions flagged by staticcheck; the oracle set was re-run after that rewrite, not assumed.

Merging.

Issue #346, opened 2026-09-21
Parser gap: a function body after the separator need not be a brace group


ss-o:
## Summary

A function body after the optional separator need not be a brace group.
zshmisc spells the definition as `word ... () [ term ] { list }`, but native Zsh accepts any statement as the body, and `functions f` confirms it reads the statement as the body rather than running it.

Both spellings are affected, so this is not specific to the `()` form fixed in #307.

## Minimized reproduction

Every row has empty `zsh -f -n` stderr (zsh 5.9.2) and is rejected by zsh-lint `main`.

| Source | zsh-lint |
| --- | --- |
| `a () ; print z` | `` `foo()` must be followed by a statement `` |
| `a () ; if true; then print z; fi` | same |
| `a () ; for i in 1; do print z; done` | same |
| `a () ; ( print z )` | same |
| `a b () ; print z` | same |
| `function a; print z` | `` `function` must be followed by a statement `` |
| `function a; if true; then print z; fi` | same |
| `a () ; { print z }` | parses (#307) |
| `function a; { print z }` | parses (#213) |

`a () ;` followed on the next line by `print z` defines `a` with `print z` as its body: the statement does not run, and `functions a` prints

```
a () {
	print z
}
```

## Classification

[Functions](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Functions), zshmisc: the body is a `list`, and a brace group is one way to spell it, not the only one.

## Corpus evidence

None in the surveyed z-shell corpus; found while verifying #307.

## Acceptance criteria

- A `gap-<id>-non-brace-function-body.zsh` fixture with the failing rows fails on the public parse path before the fix and passes `zsh -f -n`.
- Each row parses into a `FuncDecl` whose body holds the statement at its original position.
- The multi-name row restores every name, composing with #304.
- Native-invalid shapes stay rejected (`invalid-<id>-*.txt`), including `a () ;` at end of file.

## Zsh manual section

zshmisc, [Functions](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Functions).

## zsh-lint version

`main` at `67eb1a3` (with #307 applied); reproduced identically on `110324e`.

## References

- #213: `function name; { list }` adapter
- #307: separator before a `()` body brace
- #304: multi-name `()` definitions

PR #347, opened 2026-09-21
fix(parse): accept a separator before a () function body

### LINKED ISSUE #307: Parser gap: support a separator before the body in the name () ; { list } definition form
## Summary

The `name () ; { list }` spelling, a `()` definition with a separator before the body brace, is native-valid and rejected. The equivalent reserved-word spelling `function name; { list }` has parsed since #213 (`parseFunctionSemicolonBody`), but that adapter only recognises a site whose first word is `function` (`findFunctionSemicolonBody`, `internal/parse/function_compat.go`), so the `()` form falls through with the parser's own error. The multi-name form `a b () ; { : }` (#304) fails the same way after its retry.

## Minimized reproduction

All rows are `zsh -f -n` valid (zsh 5.9.2).

| Source | zsh-lint `main` (`912dbbf`) and #304 |
| --- | --- |
| `a () ; { : }` | `` 1:1: `foo()` must be followed by a statement `` |
| `a b () ; { : }` | `` 1:3: `foo()` must be followed by a statement `` (after the #304 retry) |
| `function a; { : }` | parses (#213) |
| `function a () ; { : }` | parses (#213) |

## Classification

[Functions](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Functions), zshmisc: `word ... () [ term ] { list }`; `term` is the optional separator before the body, the same `term` the `function` spelling accepts.

## Corpus evidence

None in the surveyed z-shell corpus; found while probing #251 shapes.

## Acceptance criteria

- A `gap-<id>-paren-semicolon-body.zsh` fixture with the first two rows fails on the public parse path before the fix and passes `zsh -f -n`.
- `findFunctionSemicolonBody` also recognises a site whose words are one or more plain names followed by `()`, with the same semicolon mask and the same `{` check; or a sibling gate under the same adapter. Native-invalid shapes such as `a () ; ;` and `a () ;` at end of file stay rejected (`invalid-<id>-*.txt`).
- Composes with #304: the multi-name row restores every name.

## References

- #213: `function name; { list }` adapter
- #251, #304: multi-name `()` definitions

ss-o:
Closes #307.

zshmisc spells a function definition as `word ... () [ term ] { list }`, so the separator before the body brace is optional in the `()` spelling exactly as it is after the `function` keyword.
#213 adapted the keyword form, but its scanner anchored on the literal word `function`, so the `()` form fell through with the parser's own error.

| Source | before | after |
| --- | --- | --- |
| `a () ; { : }` | rejected | parses |
| `a b () ; { : }` | rejected | parses |
| `a b c () ; { : }` | rejected | parses |
| `a () ;` + newline + `{ : }` | rejected | parses |

## Recognition without a keyword

The `()` spelling has no keyword to anchor on, so the head is confirmed by shape: one or more plain words, then `()`.
A word carrying shell syntax is not a function name, which is what keeps these valid commands out of the adapter — each means something other than a definition, and each is asserted in a test:

```zsh
(( 1 + 1 )) ; { : }
( print a ) ; { : }
x=1 ; { : }
x=(1 2) ; { : }
a | b ; { : }
a > /dev/null ; { : }
print *(.) ; { : }
```

Composes with #304: this adapter removes the separator, that one splits the names into `FuncDecl.Names`.

## Verification

- 44 sources checked against `zsh -f -n`, **zero regressions**.
- Differential over corpus + invalid fixtures + 60 real-world files: **exactly one verdict changes**, the new fixture.
- Mutation tested, 4/4 caught. One mutation I wrote first was **invalid** — `} else if false {` makes the gate unconditionally permissive rather than disabling it, so it proved nothing; replaced with one that forces `parenHeadAt` to accept everything, which fails 4 subtests.
- Tree shape asserted, not just `err == nil`: every name is checked via `FuncDecl.Names`, and the body literal is asserted to keep its original byte offset so diagnostics still point at real source.
- Two `invalid-307-*.txt` fixtures pin what Zsh rejects (`a () ;` at EOF, `a () ; ;`).
- Fixture runs under `zsh -f` (all 7 functions execute), is rejected by base, accepted after, and registered in `requiredFixtures` — proven non-vacuous by removing it.

## Scope

Verifying this surfaced a **separate, pre-existing gap**: a function body after the separator need not be a brace group at all (`a () ; print z`, and equally `function a; print z`).
It affects both spellings, predates this change, and is unchanged by it — filed as #346 rather than widened into this PR.


ss-o:
## Review

### Risk

The `()` spelling has no keyword to anchor on, so recognition is by shape — that is the whole risk surface.
A too-permissive head would claim commands that merely end in `()`, masking a separator inside something that is not a definition.

That is why `TestParenSemicolonBodyLeavesOtherCommands` asserts seven valid non-definition commands still parse untouched, and why `parenHeadAt` is tested directly rather than only through the parser, where another adapter could mask a wrong answer.

### Evidence quality

Tree shape is asserted, not `err == nil`: names are read from `FuncDecl.Names` so the multi-name composition with #304 is genuinely checked, and the body literal's byte offset is asserted unchanged so a diagnostic still points at real source.

Mutation testing caught 4/4 — but the first mutation I wrote was **invalid**, and I only found that by chasing why it "survived": `} else if false {` skips the fallback `return 0, false` and makes the gate *more* permissive rather than disabling it. The replacement forces `parenHeadAt` to accept everything and fails 4 subtests. A mutation that proves nothing is worse than none, because it reads as evidence.

### Scope discipline

Verifying this surfaced that a function body need not be a brace group at all, in **both** spellings. It is pre-existing, unchanged by this diff (confirmed base vs new on all 12 affected sources), and filed as #346 rather than folded in — the issue's own acceptance criteria assumed `a () ;` + `print z` was invalid, which is wrong; it defines the function.

Merging.

PR #348, opened 2026-09-21
fix(parse): accept a multi-name for with a parenthesized word list

### LINKED ISSUE #324: Parser gap: multi-name for with a parenthesized word list (for a b (1 2) { : }) is rejected

ss-o:
Closes #324.

zshmisc composes two productions: `for name ... [ in word ... ] term do list done` names more than one variable, and `for name ( word ... ) sublist` is the alternate form.
The front end handled each alone but not together.

| Source | before | after |
| --- | --- | --- |
| `for a b (1 2) { : }` | rejected | parses |
| `for a b (1 2 3 4) { print $a $b }` | rejected | parses |
| `for a b c (1 2 3) { : }` | rejected | parses |
| `for a b (1 2) :` | rejected | parses |

## Two body forms, two adapters

The alternate production allows a brace body and a bare sublist, and they take different paths:

- **brace body** — the multi-name `for` adapter's short-form branch now scans the extra names before the `(` and masks them, exactly as it already does for the `in` form.
- **sublist body** — `scanForSite` only moves the scan position past the names. That rewrite keeps every byte length, so nothing needs masking.

I first wrote a mask for the sublist path too. Mutation testing said removing it changed nothing, and checking directly confirmed it was **dead code** — deleted rather than left in as decoration.

## Semantics

The names bind in tuples, which the fixture proves by running rather than asserting:

```
tuple 1 2
tuple 3 4
```

The resulting tree is the same `WordIter` the `for a b in 1 2; do ... done` form produces, with names, items and body at their original byte offsets (asserted by slicing the original source at the recorded positions).

## Scope

Two rows the issue lists are **not** fixed here, because their single-name equivalents fail identically — they are body-form gaps, not multi-name gaps:

| Source | single-name twin | verdict |
| --- | --- | --- |
| `for a b (1 2); { : }` | `for a (1 2); { : }` | both rejected |
| `for a b (1 2) do :; done` | `for a (1 2) do :; done` | both rejected |

Fixing them under this issue would have meant claiming a multi-name fix for something multi-name has nothing to do with.

## Verification

- 37 sources vs `zsh -f -n`, **zero regressions**; 13 remaining disagreements are pre-existing separate gaps, each confirmed identical on base.
- Differential over corpus + invalid fixtures + 60 real-world files: **exactly one verdict changes**.
- Mutation tested. One mutation **build-failed** (proves nothing) and was rewritten to keep the variables used; one survivor turned out to be dead code and was deleted; one survivor (`>` vs `>=`) changes no verdict and is now documented as an optimisation rather than a guard.
- Adversarial cases stay correct: `for ((i=0;i<2;i++))`, `for a in (1 2)`, the select paren form, a quoted `)` inside the list, `$(...)` and globs.
- Fixture runs under `zsh -f`, is rejected by base, accepted after, registered in `requiredFixtures`, proven non-vacuous.


ss-o:
## Review

### Risk

This adapter rewrites source and rebases positions, so the failure that matters is not a missed fix but a **wrong byte mapping**: a diagnostic would then point at the wrong place in a file that parses fine.

`TestMultiNameParenForKeepsPositions` therefore asserts name, both items and the body against `strings.Index` of the original source rather than trusting the tree to look plausible.

### What mutation testing actually changed here

Three of four mutations forced a real correction rather than just scoring:

1. One **build-failed** — that proves nothing, so I rewrote it to keep the variables used and emit a zero-width mask instead; it then failed the new scanner test.
2. One survivor was **dead code**: the sublist path's mask never affected any verdict, because that rewrite preserves byte lengths. Deleted, along with the two struct fields feeding it, instead of being kept for symmetry with the brace path.
3. The last survivor (`>` vs `>=`) changes no verdict on 37 sources. It is an optimisation, not a guard, and the comment now says so rather than implying protection it does not provide.

### Scope discipline

The issue lists six failing rows; two are **not** multi-name gaps. Their single-name twins (`for a (1 2); { : }`, `for a (1 2) do :; done`) fail identically on `main`, so fixing them here would have meant shipping a body-form fix under a multi-name heading and claiming the issue was fully understood. They are left failing and named explicitly in the PR body.

Similarly `f() { for a b (1 2) { print $a }; }` still fails — verified its single-name twin fails the same way, so it is pre-existing and out of scope.

### Semantics verified by execution

The fixture runs under `zsh -f` and prints `tuple 1 2` / `tuple 3 4`, proving the tuple binding rather than asserting only that the file parses.

Merging.

PR #349, opened 2026-09-21
fix(parse): accept a function body that is not a brace group

### LINKED ISSUE #346: Parser gap: a function body after the separator need not be a brace group

ss-o:
Closes #346.

zshmisc spells a function body as a `list`, and a brace group is only one way to write one.
Native Zsh defines `a` with `print z` as its body for `a () ; print z` — the statement does not run, it *is* the body.
The front end required the body to start with `{`, so every non-brace body was rejected in both spellings.

| Source | before | after |
| --- | --- | --- |
| `a () ; print z` | rejected | parses |
| `a () ; if true; then print z; fi` | rejected | parses |
| `a b () ; print z` | rejected | parses |
| `function a; print z` | rejected | parses |
| `function a; while false; do :; done` | rejected | parses |

## Two spellings, two steps

The body gate now refuses only what cannot **open** a statement: end of input, a further separator, and the reserved words that close or continue an enclosing construct (`then`, `else`, `elif`, `fi`, `do`, `done`, `esac`, `always`). Anything else is a body.

The keyword spelling needs more than that. The upstream parser accepts `function a() print z` but **not** `function a  print z`, even though Zsh accepts both, so masking the separator to a blank is not enough. A keyword head is rewritten to carry `()`, spending the separator plus the blank after it — Zsh needs no blank between `()` and the body, so `function a; print z` becomes `function a()print z` with every later byte at its original offset.

I found this the honest way: the first implementation fixed the `()` spelling and silently left `function a; if ...` failing, and the error text (`` `foo()` must be followed by a statement `` at 1:1) pointed at the definition rather than the real cause.

## Scope

Two shapes are left failing because they are **separate pre-existing gaps**, verified by testing their twins:

| Source | why not here |
| --- | --- |
| `function a;` at EOF | Zsh accepts it and defines *nothing*; a no-op head, unrelated to bodies |
| `function a; function b; print z` | hand-written `function a() function b; print z` fails identically — nested definitions are their own limitation |

## Verification

- **47 sources** vs `zsh -f -n`, **zero regressions**; the 5 remaining disagreements are pre-existing and each confirmed identical on base.
- Differential over corpus + invalid fixtures + 60 real-world files: **exactly one verdict changes**.
- Positions asserted by slicing the original source — the `()` rewrite costs two bytes, so a wrong offset would point diagnostics at the wrong place in a file that now parses.
- Declarations asserted by name (`a b () ; print z` declares both `a` and `b`), so a rewrite that dropped the definition could not pass by parsing alone.
- Mutation tested, 7 mutations. Two survivors were **deleted as dead code** after proving unreachability (an already-parenthesized guard and a degenerate-name guard, both already covered by the keyword check); one was reclassified as an optimisation and documented as such rather than left implying protection it does not give; one survivor became caught after adding the bounds case it guards.
- Fixture runs under `zsh -f` — all 12 functions execute, `paren-multi` printing twice proves both names bind — is rejected by base, accepted after, registered in `requiredFixtures`, proven non-vacuous.


ss-o:
## Review

### The part that nearly shipped wrong

The first implementation widened the body gate and declared the issue fixed. It was not: `function a; if true; then :; fi` still failed, and the error pointed at 1:1 with `` `foo()` must be followed by a statement `` — a message that describes the definition, not the body. Only tracing the masked source showed the upstream parser rejects `function a  print z` while accepting `function a() print z`, which is why the keyword spelling needs a rewrite rather than a blanked separator.

The lesson is the one this repository keeps teaching: a partial pass over the reported shapes reads exactly like a full fix.

### Risk

This adapter rewrites source and rebases positions, so the failure that matters is a wrong byte mapping — a diagnostic pointing at the wrong place in a file that now parses. The `()` rewrite costs two bytes, so `TestFunctionNonBraceBodyKeepsPositions` asserts the body word against `strings.Index` of the *original* source, and `TestFunctionNonBraceBodyBuildsDeclaration` asserts the declared names, because a rewrite that dropped the definition would still parse and still place the word correctly.

### Mutation testing

Seven mutations, and three changed the code rather than the score:

1. An already-parenthesized guard and a degenerate-name guard both survived. Probing showed each was **unreachable** — the keyword check rejects those heads first — so both were deleted rather than pinned with tests that would have passed either way.
2. The brace test survived too, but rewriting a brace body also parses and keeps positions, so it is an **optimisation, not a guard**. The comment now says that instead of implying protection it does not provide.
3. One survivor was genuine: the bounds test on a separator at end of input. Adding that case turned it into a caught mutation.

### Scope

`function a;` at EOF and nested definitions stay failing. Both were checked against hand-written twins that fail identically on `main`, so neither is a regression and neither belongs under a body-shape heading.

Merging.

Issue #250, opened 2026-09-16
Parser gap: support a flagged second subscript with a bracket expression


ss-o:
## Summary

A second subscript that carries subscript flags and a bracket expression, `${name[a][(i)[x]]}`, still fails once #215 (second subscript) and #237 (bracket expression inside a flagged subscript pattern) both land, because each adapter's gate sees the other construct first. This is the next blocker in `z-shell/zi` `lib/zsh/autoload.zsh` after #215.

## Minimized reproduction

All rows are `zsh -f -n` valid. Results are from the #215 branch (`81628e0`), which parses `${a[b][(i)(#b)(x)]}` and `${a[b][(i)x#]}`; `${a[b][(i)[x]]}` fails with ``1:20: not a valid parameter expansion operator: `]` ``.

| Source | Result |
| --- | --- |
| `print ${a[b][(i)(#b)(x)]}` | ok |
| `print ${a[b][(i)x${s1}y]}` | ok |
| `print ${a[b][(i)[x]]}` | fails on the `]` that closes `[x]` |
| `integer idx="${functions[$s2][(i)(#b)([^[:space:]]#${s1}[^[:space:]]#)]}"` | fails the same way |

## Why the two adapters do not compose

The #215 adapter masks the `][` boundary as `, ` and retries, so the parser reads `${a[b, (i)[x]]}`; the flagged pattern then closes early at the first `]` exactly as in #237. The #237 adapter locates the pattern by scanning back from that `]` to a `name[(flags)` opener, and after the mask the flags are preceded by `, ` instead of `[`, so it returns the parser error. Running #237 first cannot help: the first error in the original source is the second `[`, which is #215's gate.

## Classification

[Array Subscripts](https://zsh.sourceforge.io/Doc/Release/Parameters.html#Array-Subscripts) and [Subscript Flags](https://zsh.sourceforge.io/Doc/Release/Parameters.html#Subscript-Flags): a subscript applied to the result of a previous subscript may itself use subscript flags, and `(i)` with a pattern is the common reverse-lookup form.

## Corpus evidence

`z-shell/zi` (`bcb20c1`) `lib/zsh/autoload.zsh:1167` and `:1172`, both `${functions[$fn][(i)(#b)([^[:space:]]#${orig_saved1}[^[:space:]]#)]}`. With #215 applied the file stops at 1167:80.

## Acceptance criteria

[...]

Issue #284, opened 2026-09-17
Parser gap: `if [[ ... ]] && { ... }; then` is read as an alternate brace-form if


ss-o:
## Summary

An ordinary `if` or `elif` whose condition list is `[[ ... ]] && { ... }` (or `(( ... )) && { ... }`) fails to parse whenever the same file also contains an alternate brace-form `if [[ ... ]] { ... }` anywhere. The alternate-if adapter takes the `{` after `&&` as the alternate body opener, and the retry then fails at that brace. This is the next blocker in `z-shell/zi` `zi.zsh` after #277.

## Minimized reproduction

All rows are `zsh -f -n` valid. Results are from `main` at `b6eb855`; the #277 branch does not change them.

| Source | Result |
| --- | --- |
| `if [[ $t == s ]] {` / `x=1` / `}` / `if [[ -n $b ]] && { true; }; then` / `:` / `fi` | `` 4:16: `&&` must be followed by a statement `` |
| the same two statements in the other order | `` 1:16: `&&` must be followed by a statement `` |
| `if (( 1 )) {` / `x=1` / `}` / `if [[ -n $b ]] && { true; }; then` / `:` / `fi` | `` 4:16: `&&` must be followed by a statement `` |
| `if [[ $t == s ]] {` / `x=1` / `}` / `if true && { true; }; then` / `:` / `fi` | parses |
| `if [[ $t == s ]] {` / `x=1` / `}` / `[[ -n $b ]] && { true; }` | parses |
| `if [[ -n $b ]] && { true; }; then` / `:` / `fi` alone | parses |
| `if [[ -n $b ]] && { print cond; } { print body3 }` | `` 1:35: statements must be separated by &, ; or a newline `` (native runs it, see Mechanism) |

Each `/` marks a newline.

## Mechanism

[...]

Issue #283, opened 2026-09-17
Parser gap: a bracket expression before `#`, `*`, or `,` in a flagged subscript pattern is cut at its `]`


ss-o:
## Summary

mvdan/sh lexes a flagged subscript pattern, `${name[(r)pattern]}`, as raw text up to the first `]`, so a bracket expression ends the pattern early (#237). The #237 adapter reopens the pattern only when the byte after that `]` makes the parser fail with one exact error; when the byte after `]` is `#`, `*`, or `,`, the parser either accepts the cut silently or fails with a different error, and the file is misread or rejected. Found while probing #277.

## Minimized reproduction

All rows are `zsh -f -n` valid. Results are from `main` at `b6eb855` and from the #277 branch, which does not change them.

| Source | Result |
| --- | --- |
| `print ${m[(r)a[^:]##]}` | parses, wrongly: the index is a `FlagsArithm` whose pattern is `a[^:`, and `##]` is read as a `##` `Expansion` operator with the word `]` |
| `print ${m[(r)[^:]##,a]}` | parses, wrongly: the pattern is `[^:`, and `##,a]` is the operator and its word |
| `print ${m[(r)a[^:]#,b]}` | parses, wrongly: the pattern is `a[^:`, and `#,b]` is a `#` operator and its word |
| `print ${m[(r)[^:],a]}` | `` 1:18: this expansion operator is a bash feature; tried parsing as zsh `` (the `,` after the cut is bash's case-modification operator) |
| `print ${m[(i)[a],3]}` | `` 1:17: this expansion operator is a bash feature; tried parsing as zsh `` |
| `print ${m[(r)[^:]*,a]}` | `` 1:18: not a valid parameter expansion operator: `*` `` |
| `print ${m[(r)[^:]]}` | parses correctly (#237) |

The `*` row reaches the #237 adapter (its error is one the adapter gates on) but `scanFlagPatternBrackets` returns false at the depth-0 `,` (`internal/parse/subscript_flag_pattern.go:193`), so the parser error stands. The two `bash feature` rows never reach it: the adapter gates on `not a valid parameter expansion operator`, `cannot be followed by a word`, and `` `a[b]` must be followed by `=` `` (`subscript_flag_pattern.go:17`), and this is a `syntax.LangError`. The three silent rows raise no error at all, so no error-gated adapter can see them.

## Mechanism

[...]

Issue #214, opened 2026-09-15
Parser gap: support foreach ... end loops


ss-o:
## Summary

The csh-style `foreach name ( word ... ) list end` form fails to parse on the pinned front end, and parses as three unrelated commands on mvdan/sh v3.14.1. The corpus already contains it.

Split from #206.

## Minimized reproduction

`zsh -f -n` valid.

| Source | zsh-lint `main` (`eb9815f`, mvdan/sh v3.13.1) | raw mvdan/sh v3.14.1 |
| --- | --- | --- |
| `foreach v ($a)` newline `cmd $v` newline `end` | ``1:15: a command can only contain words and redirects; encountered `(` `` | no error; three `CallExpr` (`foreach v ($a)`, `cmd $v`, `end`) |

`foreach v ($a); cmd $v` on one line, as written in #206, is rejected by `zsh -n`: `foreach` needs a brace body or an `end` terminator.

## Classification

[Alternate Forms For Complex Commands](https://zsh.sourceforge.io/Doc/Release/Shell-Grammar.html#Alternate-Forms-For-Complex-Commands): `foreach name ... ( word ... ) list end`, documented as another form of `for` with no option dependency; `foreach` and `end` are reserved words.

## Corpus evidence

`z-shell/zi` `tests/fixtures/public-contract/foreach/zi.zsh:25`.

## Impact

Any bump to mvdan/sh v3.14.x turns this from a loud parse error into a silent false negative. The corpus expectation for that fixture must assert the tree shape, not only that the file parses.

## Acceptance criteria

- A `gap-<id>-foreach-end.zsh` fixture fails on the public parse path before the fix; after a front-end bump it must fail on tree shape rather than pass vacuously.
- The resulting tree is a `ForClause` whose body holds the loop list, with original positions.

## References

- #206: originating report
- #209: proposes the v3.14.1 bump

PR #352, opened 2026-09-21
fix(parse): accept math function calls in arithmetic expressions

### LINKED ISSUE #233: Parser gap: support math function calls in arithmetic expressions
## Summary

A math function call inside an arithmetic expression, `name(args)`, fails to parse. The raw front end reads `(` after an identifier as an invalid operator.

## Minimized reproduction

| Source | zsh-lint `main` (`420246f`) |
| --- | --- |
| `(( x = sqrt(4) ))` | ``1:12: not a valid arithmetic operator: `(` `` |
| `print "$(( abs(-1) ))"` | ``1:15: not a valid arithmetic operator: `(` `` |
| `(( x = float(y) ))` | ``1:13: not a valid arithmetic operator: `(` `` |

## Classification

[Arithmetic Evaluation](https://zsh.sourceforge.io/Doc/Release/Arithmetic-Evaluation.html): "mathematical functions can be called with the syntax `func(args)`". `zsh/mathfunc` provides the standard set; `functions -M` defines user functions with the same call syntax. The call is a real language construct, so both `((...))` and `$((...))` are affected.

## Corpus evidence

`z-shell/zpmod` `benchmarks/run.zsh:411` (`5fab33e`): `result_stdev[$case_key]=$(printf '%.3f' "$(( sqrt(variance) ))")`.

## Resolution note

The failure is inside the front end's arithmetic parser, so a byte-preserving adapter can only mask `name(` into something the arithmetic grammar accepts; the resulting `ArithmExp` would then not represent a call. This is the second construct, after `repeat` (#208), where the adapter contract may not produce a correct tree, and it is therefore relevant to the ADR-0023 point 4 fork trigger. Decide the representation before implementing.

## Acceptance criteria

- A `gap-<id>-math-function-call.zsh` fixture covering the three rows fails before the fix.
- Nested and multi-argument calls (`max(1, min(2, 3))`) parse.
- The chosen representation is documented in `docs/project/parser-gap-workflow.md` if it uses `parse.File` metadata.

## References

- mvdan/sh#1294 (open), reference only per ADR-0023
- #208: the first construct that raised the fork question

ss-o:
Closes #233.

zshmisc, Arithmetic Evaluation, allows a math function call in an arithmetic expression.
mvdan/sh reads the identifier as a complete operand and then finds `(` where an operator would be, so every call was rejected:

```
t.zsh:1:15: not a valid arithmetic operator: `(`
```

## The first implementation was wrong, and parsing did not show it

The obvious mask turns `sqrt(4)` into the comma expression `sqrt,4` by masking the parentheses.
That parses, and every does-it-parse check passed.

But a comma binds loosest of all arithmetic operators, so it captures the neighbours:

| source | tree it produced |
| --- | --- |
| `$(( sqrt(4) + 1 ))` | `(sqrt , (4 + 1))` |
| `$(( 1 + sqrt(4) ))` | `((1 + sqrt) , 4)` |

The name became a variable and the argument escaped the call.
A rule reading that tree would be wrong about the arithmetic.

Masking the **name** instead and keeping the parentheses gives a self-contained operand that binds where the call did:

| source | tree |
| --- | --- |
| `$(( sqrt(4) + 1 ))` | `([4] + 1)` |
| `$(( 1 + sqrt(4) ))` | `(1 + [4])` |
| `$(( sqrt(4) * 2 + 3 ))` | `(([4] * 2) + 3)` |

`TestMathFunctionCallKeepsPrecedence` pins these shapes so the regression cannot return silently.
The corpus fixture proves it at runtime too: it prints `precedence 5` and `precedence-mul 11`, which under the comma mask would have been `17` and `35`.

A call with no arguments has nothing to keep between its parentheses, so it becomes a literal `0` padded to the original width.

## Metadata

mvdan/sh has no call node in its arithmetic grammar, so each call's name and arguments are retained as `File.MathFunctionCalls` with their original source positions, alongside the existing second-subscript and repeat-loop metadata.
Without this the name is simply gone, and a rule could not tell `sqrt(4)` from the grouping `(4)`.

## Scope

A name must touch its `(`.
Native Zsh rejects `sqrt (4)` with `bad math expression: operator expected`, so the blank form stays rejected, as does a digit-led name such as `4(2)`.
Parentheses outside an arithmetic context — subshells, array literals, glob qualifiers, `case` patterns, definitions — are never scanned.

## Verification

- **9 of 12** probe shapes fixed; the 3 still rejected are correct: `sqrt (4)` and `1 +* 2` are Zsh runtime errors, and `${a[int(2.7)]}` is a different error belonging to the subscript family (#250).
- **33 sources** checked against `zsh -f -n`: **zero regressions**. The 5 disagreements are each identical on base.
- **18 adversarial non-call parenthesis shapes**: zero verdict changes.
- Differential over the corpus and 60 external sources: **exactly one** verdict change, this fixture.
- **48 subtests**; mutation testing **6/6 caught**, including the parenthesis mask that produced the precedence bug above.
- `gofmt`, `go vet`, full suite, `-race -count=2` clean. Fixture guard confirmed non-vacuous.

## Impact

This closes the last parser gap in `z-shell/zpmod`: all **58** of its Zsh sources now parse.
It was found by re-running discovery over `zi` and `zpmod`, whose last recorded report predates roughly a dozen merged gap fixes.


ss-o:
## Review

Self-review under the ADR-0026 fallback (Copilot rate-limited).
GitHub refuses an approving review on one's own pull request, so this comment carries the review record; the maintainer authorization for this change is already granted.

**Correctness.** The load-bearing question was not whether the call parses but what tree it produces. The first implementation masked the parentheses into a comma expression; it parsed everything and was wrong, because a comma binds loosest and pulled the neighbouring operands into the call (`1 + sqrt(4)` → `(1 + sqrt) , 4`). Masking the name and keeping the parentheses gives an operand that binds where the call did. Four expression shapes are asserted structurally, and the corpus fixture cross-checks the same arithmetic by execution (`precedence 5`, `precedence-mul 11`) — values that would differ under the rejected approach.

**Scope.** Scanning is confined to `$(( ))` and `(( ))` interiors, so a parenthesis that opens a subshell, array literal, glob qualifier, `case` pattern or definition is never masked; 18 such shapes are pinned as unchanged. The name-touches-paren rule and the digit-led-name rule both match native Zsh, which rejects `sqrt (4)` and `4(2)`.

**Metadata.** `File.MathFunctionCalls` keeps each name and argument list at original source offsets, following the existing second-subscript precedent. Without it the masked name would be unrecoverable and a rule could not distinguish `sqrt(4)` from `(4)`.

**Evidence.** 48 subtests; 6/6 mutations caught, including the exact parenthesis-masking mutation that reproduces the precedence bug; 33 sources against `zsh -f -n` with zero regressions and every disagreement confirmed identical on base; exactly one verdict change across the corpus and 60 external sources; `-race -count=2` clean; fixture guard confirmed non-vacuous by removal; CI 14/14.

**Residual risk.** The scanner is byte-based, consistent with the other adapters. Quoted text inside arithmetic is skipped, and command substitution inside arithmetic is unaffected — both covered by tests. Remaining `zi` gaps belong to other filed issues (#250, #283, #284, #214).

Approving on the evidence above and merging.

PR #353, opened 2026-09-21
fix(parse): keep a condition brace group out of the alternate-if body

### LINKED ISSUE #284: Parser gap: `if [[ ... ]] && { ... }; then` is read as an alternate brace-form if

ss-o:
Closes #284.

An ordinary `if ... ; then` whose condition ends in a brace group failed to parse whenever the same file also held an alternate brace-form construct:

```zsh
if [[ -n $t ]] {
  x=1
}
if [[ -n $b ]] && { true; }; then
  :
fi
```

```
t.zsh:4:18: `;` can only immediately follow a statement
```

## Cause

`scanAlternateConditionBrace` returned the first `{` it found as the body opener, including one written directly after `&&` or `||`.
The adapter then rewrote that condition brace into an alternate-form body.

Alternate Forms For Complex Commands requires the test to be "suitably delimited, such as by `[[ ... ]]` or `(( ... ))`, else the end of the test will not be recognized", so a `{ ... }` after a connective is part of the condition list.
Native Zsh agrees, and the three cases are distinguishable by execution:

| source | `zsh -f` |
| --- | --- |
| `if [[ -n $b ]] && { print body }` | parse error: the group is the condition, so there is no body |
| `if [[ -n $b ]] && [[ -n $c ]] { print body }` | runs: a delimited test re-arms the body |
| `if [[ -n $b ]] && { print cond; } { print body }` | runs, printing **both**: group is condition, second brace is body |

The bug needed two constructs in one file because the adapter only runs when a file has an alternate form to fix — which is exactly why this reached `zi.zsh` rather than being caught earlier.

## Fix

The scan tracks whether it stands just past a delimited test.
A connective clears that, so a brace after `&&`/`||` is skipped as a condition group instead of returned; the scan continues past its `}`, where a further delimited test — or the group itself — re-arms the body.

That last part also fixes the third row above, so the issue's "or record it as a separate gap" clause is not needed.

## Scope

Seven shapes fixed: `if`, `elif`, `while`, `||`, a negated group, two groups in one condition, and condition-group-then-body.
`if [[ -n $b ]] && { print body }` stays rejected, matching Zsh, and is pinned by a test and an invalid fixture.

## Verification

- **34 sources** against `zsh -f -n`: **zero regressions**. One disagreement, `alt_multiline`, is identical on base — a pre-existing multi-line alternate-condition gap unrelated to this change.
- **20 adversarial shapes**: no regressions, no hangs; three of them turned ERR→OK, all genuine fixes.
- Differential over the corpus and 60 external sources: **exactly one** verdict change, this fixture.
- **86 subtests**. Tree shape asserted, not just `err == nil`: the condition must be `BinaryCmd(&&){TestClause, Block}`, and for the condition-then-body row the body text must be `print BODY`.
- Mutation testing **6/6 resolved**: 5 caught; the sixth changes no observable behaviour (the `default` branch can never return a `{` offset, and the caller acts only on a `{`) and is documented as inert rather than dressed up as a guard. One mutation initially failed to compile — rerun so it proved something.
- The fixture executes under `zsh -f`: every branch prints, including `condition-part` **and** `body-part`.
- `gofmt`, `go vet`, full suite, `-race -count=2` clean; fixture guard confirmed non-vacuous.

## Impact

`zi.zsh` advances from `2598:52` to `2749:44` — a multi-line arithmetic ternary, a separate gap not currently filed. I will file it next.
`zi` is now at 4 remaining gap files (#250, #283, #214, and the new ternary); `zpmod` stays at 0.


ss-o:
## Review

Self-review under the ADR-0026 fallback (Copilot rate-limited).
GitHub refuses an approving review on one's own pull request, so this comment carries the review record; the maintainer authorization for this change is already granted.

**Correctness.** The rule implemented is Zsh's own: a body brace must follow a *delimited* test. Rather than infer that from the grammar text alone, all three cases were separated by execution under `zsh -f` — the no-body form is a parse error, the delimited-test form runs, and the condition-group-then-body form runs printing both words. The implementation follows that directly: a connective clears the "just past a delimited test" state, and the scan continues past a condition group's `}` so a later brace can still be the body.

**Scope.** This narrows when a brace is treated as a body, so the risk is suppressing a real alternate-form body. Nine alternate-form shapes are pinned as still recognized, and the differential over the corpus plus 60 external sources shows exactly one verdict change. Three adversarial shapes moved ERR→OK; each is `zsh -f -n` valid, so they are fixes, not laxity.

**Shape, not just acceptance.** A fix that moved the condition group into the body would still parse. The tests assert the condition is `BinaryCmd(&&){TestClause, Block}`, and for the condition-then-body row that the body text is exactly `print BODY` — matching what `zsh -f` executes.

**Mutation testing.** 6 mutations, 5 caught. The sixth (`default: return i` → `return len(src)`) survived; measuring it showed that branch can never return a `{` offset and the caller acts only on a `{`, so the two are observationally identical. It is documented as inert rather than defended with a test that would pass either way. One mutation initially failed to compile — a build failure proves nothing, so it was rewritten to compile and then caught.

**Termination.** Two new `continue` paths were added, so non-advancing loops were checked explicitly with a timeout probe, and every survey run in this PR used `timeout` with an explicit HANG verdict. No hangs.

**Residual risk.** `alt_multiline` (a multi-line alternate-form condition) still fails, identically on base — pre-existing and unrelated. The next `zi.zsh` error, `2749:44`, is **not** a ternary gap as it first appeared: it is a math function call in a ternary's true-branch, where the #233 adapter's error gate does not list `ternary operator missing ':' after '?'`. That is a gap in the gate merged as #352, and I am filing and fixing it separately rather than widening this PR.

Approving on the evidence above and merging.

